PR #298, opened 2015-03-29
More complete support for window padding

m4rw3r:
I have been using a patched version I made of https://github.com/ianyh/Amethyst/pull/256 which (in a really hacky way) added border-collapse so that all padding is uniform.

Some time ago I saw that few patches fixed problems with switching spaces, so of course I wanted that functionality too, but I didn't want to give up window padding. Sadly the pull-request couldn't just be moved, got some conflicts, so I decided to reimplement window-padding from scratch, and this time including settings for the padding amount as well as for padding-collapse.

This is the first time I have used the interface-builder to attach some GUI-elements, and it seems like the changes I made affected the resulting XML in many places where it shouldn't need to. So I need some feedback and/or help to clean this up and then I can squash it all into a single neat commit.

I know of at least one issue: it does not trigger changes when updating the settings, windows need to be resized manually before the new padding is applied or Amethyst needs to be restarted.


leonardt:
Cool testing this out now, looks good so far

leonardt:
This has been working well for me so far

Amethyst/default.amethyst
```diff
@@ -215,5 +215,7 @@
     "mouse-follows-focus": false,
     "focus-follows-mouse": false,
     "enables-layout-hud": true,
-    "enables-layout-hud-on-space-change": true
+    "enables-layout-hud-on-space-change": true,
+    "window-padding": 10,
```
ianyh: Shouldn't have any window padding by default.

ianyh:
This runs into a couple problems:
* Has weird behavior for small frame changes. There is a 20px leeway in frame size where we consider a frame to not have changed if its within that error. This leads to particularly bad behavior with a padding of 10px.
* Has weird behavior for windows that are pushed off screen because they have a minimum frame. They touch the screen edges and it ends up looking pretty weird.

leonardt:
What's the reasoning behind the 20px leeway?

ianyh:
Terminal only sizes windows in units of cursor size, so if you don't give it some leeway you infinitely try to change it's frame which causes really weird flickering. At least that was true the last time I tested it. That may have actually been two major OS X releases ago. I should test it again.

ianyh:
Hm. I can't reproduce the flickering anymore. I'll test the padding without the leeway.

m4rw3r:
> Has weird behavior for windows that are pushed off screen because they have a minimum frame. They touch the screen edges and it ends up looking pretty weird.

I don't know if there is any way to avoid this, as the alternative is to have it not get pushed off screen but instead intersect with the other programs, which will end up looking (imho) worse. No matter what we do windows which have a minimum size will look odd; eg. even without padding windows with minimum height can look odd when placed in a slot which forces them to extend downwards into the window placed below (Skype seems to be a common culprit of this on my machine).

I did find one more issue with this though, when windows extending off-screen are moved back, they are only moved back on screen so much as to be visible, and do not respect the padding. I do not know if this is a good thing or a bad thing, considering low-resolution screens could get cramped if windows are pushed too much towards the center of the screen (or worst case scenario is that the window ends up being outside the screen anyway, pushed outside by the padding).

> Hm. I can't reproduce the flickering anymore. I'll test the padding without the leeway.

Also try with other programs, eg. MacVim has the same behaviour as the OS X terminal and will resize to multiples of its line-height.

ianyh:
I was able to reproduce the flickering, unfortunately. I also can't get the resizing to work correctly. I keep getting weird behavior like:

screen shot 2015-04-11 at 10 19 19 am

m4rw3r:
> I was able to reproduce the flickering, unfortunately. I also can't get the resizing to work correctly. I keep getting weird behavior like:

I have not managed to reproduce it with arbitrary windows, the only time when they are not aligned properly for me is if the window itself does not let Amethyst resize it to the pixel (eg. Terminal, MacVim) or when a window has a minimum size which makes it disregard width and height (eg. Skype).

> This leads to particularly bad behavior with a padding of 10px.

Can you attempt to detail this further? What kind of bad behaviour? The only thing I can see is that Amethyst pretty much refuses to apply the padding initially when turning on padding, it is solved by resizing and moving the window so that it is outside the leeway.

So to summarize the problems this has so far:
* Does not trigger changes when updating the settings, windows need to be resized manually before the new padding is applied or Amethyst needs to be restarted (partially the need to mark windows as needing reflows, partially the leeway in resizes).
* Windows with a minimum size look odd when assigned an area which is smaller than the minimum size, as the window extends into the padding.

Is that correct?

For the last one I do not know of any suitable solution. And this is also a problem when using lots of windows on eg. the Tall or Wide layout, as they will start to overlap. Doesn't look as bad, maybe, but it is still the same problem.

maus-:
Has this been resolved? Really looking forward to this feature.

m4rw3r:
I am still using an old patched version, the same as in this pull-request. So far I have still not managed to get any flickering or windows which keep getting resized.

The issue with windows with a minimum size which is smaller than the assigned area is still a problem, but this is not possible to resolve within the constraints of OS X's window manager as far as I know (the way to go would probably be to partially hide the parts extending outside of the assigned area if they are unfocused).

At the moment I am using the native resolution on my 15" MacBook Pro which minimizes the problem of the latter, it is very rarely I get a window which has a large enough minimum size.

ianyh:
I think I would be fine with putting this feature in as long as it is explicitly marked as experimental. It's the kind of thing that I can imagine running into a lot of edge cases. I can deal with reorganizing the preference pane to make experimental options explicitly marked as such. I have to do it for focus-follows-mouse anyway.

Regardless, this has conflicts that need to be resolved before I can merge.

leonardt:
@m4rw3r any interest in resolving these conflicts? Alternatively you could give me access to your fork and I can do the resolution

m4rw3r:
@leonardt I am going to try to get it done this weekend

m4rw3r:
@leonardt Updated to latest master, so far seems to work well for me

@ianyh Needs some adjustments for the preferences pane, and this issue still remains:

* Does not trigger changes when updating the settings, windows need to be resized manually before the new padding is applied or Amethyst needs to be restarted (partially the need to mark windows as needing reflows, partially the leeway in resizes).

ianyh:
Alright, going to merge this for now. Can you open an issue for the padding change thing?

jclusso:
Is this going to be released soon? Very excited for this :)

jeromenerf:
It has been 399 days @ianyh since https://news.ycombinator.com/item?id=8770192 , thanks for the commitment :)

ianyh:
Harsh.

maus-:
Not super sympathetic as I've seen similar PR's with similar features merged but I never see this ever enabled in any of the 'beta builds'

jeromenerf:
@ianyh @maus-  no offense intended, awkward pun maybe :dancers:

Thanks a lot for the good work.

maus-:
It's all love, but I've been waiting for a release with padding for so long ;_;

jeromenerf:
BTW, "margin" or "spacing" may be a better fit than "padding" since the feature is about adding empty space around windows and not within. 

A few years ago, I came up with "useless gaps". Not very clever.

I have seen some iterm forks with increased "internal" padding settings.

jclusso:
I'm waiting so much for this. I literally check for updates everytime hoping this will be in it :)

ianyh:
I was waiting on doing a 0.10 to get better testing infrastructure, but maybe I should push something out sooner rather than later to get this out. I didn't realize it had been that long. Life, family, work, illness, etc.

maus-:
Do it! I totally understand this is an experimental feature and if I can contribute back with bugs or fixes I'd be happy to. Plus then I'd get my precious window margin. :D

jeromenerf:
@ianyh do as you feel. You can see you have beta testers and happy users anyway.

ianyh:
Alright, pushed out a 0.10 beta with window margins (renamed it). Let me know how it is and file issues for any bugs!

jclusso:
This is awesome! Will report back with any bugs I find.

jeromenerf:
Thanks @ianyh

jeromenerf:
@ianyh @m4rw3r as far as I can tell, it works nicely with all layouts after a relaunch.

m4rw3r:
I started using the beta yesterday (yay no more having to compile my own), and so far it all seems to be working well. Though I have noticed that it sometimes the layout hud pops up even when I am not switching programs or spaces (using 0.10b2 (43) at the moment).

Good thing that collapsed margins is the default since I personally prefer to have the same spacing everywhere (not sure where this configuration option went?).

PR #374, opened 2015-12-21
Fix for titleless windows -- AXDialog windows

ddelnano:
This is still a WIP but opening to get some help from other people.

So I have been able to fix the issue where title-less windows were ignored by Amethyst.  The subrole that was needed was `AXDialog`.  The commit I added now makes Amethyst aware it needs to tile these windows.  A working version of this can be see below.  A blank space is shown because I used quicktime to record the video on my screen and it is also an `AXDialog` window.

ezgif com-video-to-gif

All shortcuts work as expected except for ones that deal with focus.  I will continue looking into the focus thing further but since I don't know much about the project I didn't know if someone would know where this additional change needs to be made.

### DIFF
--- a/Amethyst/SIWindow+Amethyst.m
+++ b/Amethyst/SIWindow+Amethyst.m
@@ -25,6 +25,7 @@ static void *SIWindowFloatingKey = &SIWindowFloatingKey;
 
     if (!subrole) return YES;
     if ([subrole isEqualToString:(__bridge NSString *)kAXStandardWindowSubrole]) return YES;
+    if ([subrole isEqualToString:(__bridge NSString *)kAXDialogSubrole]) return YES;
 
     return NO;
 }


ianyh:
I think there was a reason I avoided using the dialog subrole, but I can't remember off the top of my head why. I'll do some digging when I get a chance.

ddelnano:
Ok.  After some testing, the focus switching does work.  It just does not allow switching focus between windows of the same app.

ddelnano:
Did you get the chance to look at this further?

koekeishiya:
I had to handle this case as well when working on Kwm. The reason for not allowing AXDialog windows by default, is that applications such as Safari (and Finder) among others use this type for popup windows or ui-elements.

ianyh:
That definitely sounds familiar. Thanks @koekeishiya. Sorry for going AWOL @ddelnano. Possibly we should just special-case iTerm or something.

ddelnano:
Thanks for that @koekeishiya. 

Maybe a white list could be implemented. That way for other applications that would have similar issues it could behave as expected?  Again I don't really know much about the code base but just a suggestion. I can maybe take a better look when I have more free time.

koekeishiya:
I have no idea how Amethyst works internally, so I'm not sure if this approach would work.
However, I solved this issue by allowing users to specify additional allowed roles that are application specific. This way you can allow iTerm windows with the specified role to tile. Granted they would have to know how to work with the Accessibility Inspector provided by OSX to use this.

ddelnano:
Wait so your saying I could get iTerm2 to work with Amethyst's current state by using the Accessibility Inspector?

ianyh:
I believe he is saying that you could use the Accessibility Inspector to figure out what parameters you could use to narrow down the logic to allow iTerm windows. It would still require modifying the code.

koekeishiya:
@ianyh Yeah that is more or less what I'm trying to say. I'm just really bad at explaining things.
So basically I imagine Amethyst reads a config, and if a user notices that some window does not get tiled by Amethyst, he could use the Accessibility Inspector to figure out which role that window has.

Then he could add to his config something along the lines of {"iTerm", AXDialog}. This would cause Amethyst to check these roles for iTerm and allow them, but not for other applications.

ddelnano:
Ah ok I misunderstood what you were saying. I thought you said it was something that could already be done. But what you described is the exactly what I was getting at with the white list haha

ddelnano:
Will try to take a look at how that could be implemented this weekend when I have some free time.

ddelnano:
Since this discussion I have been using tmux to satisfy my needs of multiple terminal windows and since it is cross platform I will be sticking with it.  So I will not pursue this addition.

PR #397, opened 2016-02-06
Minor Development Setup

interstateone:
I pulled down Amethyst to try adding a new feature but ran in to a few things that needed to be done before I got started. Hopefully these changes are necessary and aren't something that I just missed reading somewhere else :smile:.

### DIFF
--- a/Amethyst/AMShortcutsPreferencesViewController.h
+++ b/Amethyst/AMShortcutsPreferencesViewController.h
@@ -7,7 +7,6 @@
 //
 
 #import <Cocoa/Cocoa.h>
-#import <MASPreferences/MASPreferencesViewController.h>
 
 @interface AMShortcutsPreferencesViewController : NSViewController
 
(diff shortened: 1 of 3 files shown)


ianyh:
Oh no! I haven't used the `Rakefile` in so long I didn't notice it was wrong. It should actually be running `carthage fetch && carthage build` or something like that. The Sparkle update is unexpected. :(

interstateone:
Okay, let me know what in particular you'd like to change it to, or if you'd prefer to handle the changes on your own. Other than those few particulars I think it would be good to have some sort of "do this to get started building Amethyst" in the README.

ianyh:
The change to the README is great. I'll update the Rakefile when I get a chance and then merge this after that.

ianyh:
Okay, I updated the dev setup stuff so this can get merged now. I ended up updating Sparkle anyway.

PR #400, opened 2016-02-10
Updates used ruby version in travis

deiga:


deiga:
Hold with this, the travis build still is missing some finer details

deiga:
Now this actually makes the build run on travis. There's still some weird issue with importing ReactiveCocoa. Any help?

ianyh:
I'm actually without a computer temporarily (in for repairs), but I will try to check with this soon.

.travis.yml
```diff
@@ -7,4 +10,6 @@ before_install:
         - brew upgrade carthage || true
         - gem install cocoapods --pre || true
         - export LC_ALL="en_US.UTF-8"
-script: rake test
+script:
+  - carthage update --no-use-binaries
```
ianyh: This should be `carthage bootstrap --platform OSX`. Using binaries is fine.

ianyh:
The Rakefile is now out of date and I have one comment that might fix things.

.travis.yml
deiga: the `--no-use-binaries` is a fix for hitting the GitHub API limit all the time. And yeah, I'll update that to bootstrap

deiga:
I've merged the Rakefile changes, Travis build still fails because: `/AMPrefixImport.h:13:9: fatal error: 'ReactiveCocoa/ReactiveCocoa.h' file not found`

Maybe the the Carthage modules are not added to the classpath properly?

deiga:
@ianyh Any more help with this? This import error is really frustrating..

ianyh:
Take a look at my PR for regenerating the test target. I'm getting new and exciting errors over there. Can you try working off of that and see what you find?

ianyh:
For some brief background: the test target was actually too out of date to save so I just killed it.

deiga:
It works!

ianyh:
Yay! Thanks so much for digging into this.

Issue #164, opened 2014-04-19
Move window to next/prev space


leonardt:
Space IDs get a bit confusing to remember with multiple spaces and multiple monitors.  Would be nice to be able to just move a window to the next/previous space.

[Trello Card](https://trello.com/c/T5ZkdNFZ/52-amethyst-move-window-to-next-prev-space)

adbl:
+1 !

metalseargolid:
+1

nodar-chkuaselidze:
+1

Pitometsu:
I see https://github.com/ianyh/Amethyst/pull/458 implementation, but there no keys for it in preferences. Is it deprecated feature. It would be pretty useful. Could you reopen this feature request, please, if current version miss this functionality. AFAIK SizeUp have such functionality, so it should be possible.

ianyh:
It looks like it is incorrectly left out of the in-app preferences. It is `mod2+left` and `mod2+right` by default. I'll open a separate issue to address them not being in preferences.

ianyh:
Opened #660

PR #401, opened 2016-02-10
Add shortcuts to move focused window left/right a space

beneills:
# Shortcuts #

This changeset adds two shortcuts:
 * `mod2 + left`, and
 * `mod2 + right`

These act in a similar way to the existing `mod2 + [n]`, but in a relative, not absolute, fashion.

The behaviour mirrors symmetrically the OSX default `ctrl + [n]` vs. `ctrl + left`/`ctrl + right`
for space switching. The new keybindings are unlikely to clash, fit logically with existing
Amethyst/System keybindings, are idiomatic, and natural to use.

# Implementation #

The Spaces API seems to be undocumented/reverse-engineered, so I had to copy existing Amethyst
code.

I added four convenience functions for getting the current space, and for getting the number
of configured spaces.  These can be supplied a screen, or try to find the "currently focused" one.

Otherwise, I followed the implementation of `pushFocusedWindowToSpace`, and updated the README.

N.B. I return -1 from certain functions to signal failure.  Is this idiomatic Objective C?
N.B. I freely cast between `NSUInteger` and `CGSSpace`.  Is this acceptable?

# Testing #

I am unable to test the code, so __please compile and run the code before merging it anywhere
important!__

[Trello Card](https://trello.com/c/I4UPh7QB/198-amethyst-add-shortcuts-to-move-focused-window-left-right-a-space)

### DIFF
--- a/Amethyst/AMConfiguration.m
+++ b/Amethyst/AMConfiguration.m
@@ -58,6 +58,8 @@ static NSString *const AMConfigurationCommandSwapCCWKey = @"swap-ccw";
 static NSString *const AMConfigurationCommandSwapCWKey = @"swap-cw";
 static NSString *const AMConfigurationCommandSwapMainKey = @"swap-main";
 static NSString *const AMConfigurationCommandThrowSpacePrefixKey = @"throw-space";
+static NSString *const AMConfigurationCommandThrowSpaceLeftKey = @"throw-space-left";
+static NSString *const AMConfigurationCommandThrowSpaceRightKey = @"throw-space-right";
 static NSString *const AMConfigurationCommandFocusScreenPrefixKey = @"focus-screen";
 static NSString *const AMConfigurationCommandThrowScreenPrefixKey = @"throw-screen";
 static NSString *const AMConfigurationCommandToggleFloatKey = @"toggle-float";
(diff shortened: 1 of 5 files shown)


ianyh:
This is awesome. I'll try to review soon.

ianyh:
This looks good, but the command needs to be added to the preferences window. You just need to add it to the dictionary returned in `[AMConfiguration hotKeyNameToDefaultsKey]`.

ianyh:
Tracking the existing issue #164

cormacrelf:
Any reason why we wouldn't want to fill out the `hjkl` keys and assign `mod2 + H/L` by default? I've been using those combos in a BetterTouchTool mapping to do exactly this, seems logical enough and you get to stay near the rest of the default bindings, none of which require pulling your hand away from the main keys.

ianyh:
I think the logic behind the `left`/`right` thing is that it is in parallel with the system shortcuts for switching between workspaces. With the current implementation the amethyst commands are fairly inextricably tied to the system ones. Keeping things on the home row is pretty compelling, though. Will have to think harder about what the defaults are.

ianyh:
@beneills I'm going to take this over and convert to swift.

beneills:
@ianyh Great!

ianyh:
Replaced by #458. Kept your commit, though, so you still get credit for the implementation! Thanks for the contribution!

PR #477, opened 2016-08-06
fix platform for carthage

glibsm:
OSX is not a valid value. (was it renamed?)

```
$ carthage bootstrap --platform OSX
Invalid value for '--platform': OSX
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,5 @@
 task :setup do
-  system 'carthage bootstrap --platform OSX' or exit!(1)
+  system 'carthage bootstrap --platform Mac' or exit!(1)
   system 'rbenv install -s' or exit!(1)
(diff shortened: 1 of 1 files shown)


ianyh:
What version of Carthage?

ianyh:
Oh, I thought this was an issue, not a pull request. This seems fine to me! Thanks for the change!

glibsm:
Double checked my version of carthage, looks like I was a few years behind :)

`--platform OSX` now works with the new versions, but is't just an alias for `Mac`, so there is no harm that this has landed.

Looking forward to trying to contribute more in the future.

ianyh:
Yeah. Particularly given the macOS name change, using mac instead of osx seems more correct anyway.

Issue #197, opened 2014-08-11
Changing the step size?


girvo:
Hey there,

Is it possible to change the "step size" for growing/shrinking layouts? I recently got a new 28" 4K monitor, and the step size is just a little too large to have the finesse I'm after :)

Cheers!

[Trello Card](https://trello.com/c/4Z3FbVhp/66-amethyst-changing-the-step-size)

ianyh:
Closed by #479

PR #479, opened 2016-08-07
Configurable step size

glibsm:
Allows to change the default 5% jump size to something else.

`size change step` is probably a confusing name for it on the preference pane, I'm open to suggestions of how to name it better.

The unit is hardcoded to precent. I think in the future it might be interesting to allow people to change the units from % to pt to allow more flexibility. For example, they could set `2%` or `100px` steps.

After playing with this for a while, I've found that 2.5% suits me much better as the default than 5.

image

[Trello Card](https://trello.com/c/hst8KE2y/227-amethyst-configurable-step-size)

### DIFF
--- a/Amethyst/Layout/MiddleWideLayout.swift
+++ b/Amethyst/Layout/MiddleWideLayout.swift
@@ -90,17 +90,17 @@ public class MiddleWideLayout: Layout {
     override public class var layoutName: String { return "Middle Wide" }
     override public class var layoutKey: String { return "middle-wide" }
 
-    private var mainPaneRatio: Double = 0.5
+    private var mainPaneRatio: CGFloat = 0.5
 
     override public func reflowOperationForScreen(screen: NSScreen, withWindows windows: [SIWindow]) -> ReflowOperation {
         return MiddleWideReflowOperation(screen: screen, windows: windows, layout: self, windowActivityCache: windowActivityCache)
     }
 
     override public func expandMainPane() {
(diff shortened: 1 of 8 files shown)


glibsm:
#197

ianyh:
This is awesome! 🎉

You are correct that the name is a little confusing. I'll ponder on options there.

glibsm:
@ianyh any thoughts on this?

glibsm:
perhaps `window resize step`?

ianyh:
Maybe the word interval works better than step. `window resize interval`?

glibsm:
@ianyh I'm cool with that, I'll change the wording.
Rest of the diff LGTY?

glibsm:
@ianyh any thoughts on this?

ianyh:
This still looks good. I was waiting for a release to happen before merging this in, but I've been really sick and haven't been able to do much. I might cut a branch for a release and then merge this in, but not sure when.

rdewolff:
it's not clear for me what this "Window Resize Interval" means

glibsm:
> it's not clear for me what this "Window Resize Interval" means

Sorry about that. Are you confused about the functionality or the naming?

If functionality, this lets you shift the divide between the windows to give a particular section more space.

If the naming, then please feel free to suggest something else. As you can see in the PR we weren't able to come up with a better name.

rdewolff:
The functionality is not clear to me.

PR #497, opened 2016-10-27
Main pane adjustment in Row layout

glibsm:
Allows the main pane shrink/expand in row layout. Very useful for things like a large web browser and a small terminal/mail/chat client at the bottom. Otherwise I've found Row layout to be a little tricky to use.

Still a lot of drying up can be done in the layout class, but that's to come in a little bit.

Screenshot in action:
screenshot

[Trello Card](https://trello.com/c/4Bv1oOGE/240-amethyst-main-pane-adjustment-in-row-layout)

### DIFF
--- a/Amethyst/Layout/RowLayout.swift
+++ b/Amethyst/Layout/RowLayout.swift
@@ -70,7 +70,6 @@ private class RowReflowOperation: ReflowOperation {
     }
 }
 
-// TODO(glib): there is quite a lot of this copied around between Layouts... probably should abstract
 public class RowLayout: Layout {
     override public class var layoutName: String { return "Row" }
     override public class var layoutKey: String { return "row" }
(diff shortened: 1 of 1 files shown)


ianyh:
Are you sure you're not looking for the Wide layout?

glibsm:
Huh. I guess I was... why is it called wide though? :)

glibsm:
This row and wide differ in how they split the remaining space thought... wide does side to side, this still stacks vertical.

glibsm:
For example this is not possible with Wide:

screen shot 2016-10-27 at 11 56 54 am

glibsm:
@ianyh any thoughts on this? Obviously there is now a merge conflict that needs fixing, but do you think this functionality is useful?

ianyh:
Sorry for the delay. This seems like a useful addition. We might want to consider doing this for the column layout as well.

ianyh:
Per #580, closing.

PR #534, opened 2017-01-04
Auto Binary Space Partition

djp3:
This adds an auto binary space partition ("absp")

When a new window appears, Amethyst will find whichever existing window has the most screen area and split its area into two equal halves.  One half will be given to the new window and one half to the window that was previously in that space.

 "auto" in this case means that the focus window doesn't have any bearing on the split like it does in the binary split partition scheme.

The window with the most area is split along the longest axis so that windows are biased toward being square.

absp also cleanly re-lays out the windows when the windows are closed so that all leftover space is reclaimed by the remaining windows.  If windows are added and removed in LIFO order then the layout will remain stable.

If a user moves a window and then forces a relayout, then the window that is moved will drop into the closest location.  Ideally the relayout would happen when the window stops being moved, but getting that callback from OSX seems non-trivial.  (At least I couldn't figure it out)

It also does a biased split if you use the hot key for increasing the main pane size.  Initially it's a binary split, but if you increase it the ratio changes.  

Here's a zipped up animation of the behavior:
[Terminal_ScreenShot_001.mp4.zip](https://github.com/ianyh/Amethyst/files/684107/Terminal_ScreenShot_001.mp4.zip)

terminal_screenshot_002

[Trello Card](https://trello.com/c/jYHChwRt/251-amethyst-auto-binary-space-partition)

### DIFF
--- a/Amethyst/Layout/ReflowOperation.swift
+++ b/Amethyst/Layout/ReflowOperation.swift
@@ -62,150 +62,157 @@ open class ReflowOperation: Operation {
         }
     }
 
-    open func isDegenerateWindow(window:SIWindow) -> Bool{
+    open func isDegenerateWindow(window:SIWindow) -> Bool {
         let window_frame = window.frame()
         if window_frame.origin.x < 0.0 {
             return true
         }
-        if window_frame.origin.x.isInfinite{
+        if window_frame.origin.x.isInfinite {
             return true
         }
-        if window_frame.origin.x.isNaN{
+        if window_frame.origin.x.isNaN {
             return true
         }
         if window_frame.origin.y < 0.0 {
             return true
         }
-        if window_frame.origin.y.isInfinite{
+        if window_frame.origin.y.isInfinite {
             return true
         }
-        if window_frame.origin.y.isNaN{
+        if window_frame.origin.y.isNaN {
             return true
         }
 
         if window_frame.width <= 0.0 {
             return true
         }
-        if window_frame.width.isInfinite{
+        if window_frame.width.isInfinite {
             return true
         }
-        if window_frame.width.isNaN{
+        if window_frame.width.isNaN {
             return true
         }
         if window_frame.height <= 0.0 {
             return true
         }
-        if window_frame.height.isInfinite{
+        if window_frame.height.isInfinite {
             return true
         }
-        if window_frame.height.isNaN{
+        if window_frame.height.isNaN {
             return true
         }
         return false
     }
-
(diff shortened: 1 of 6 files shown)


varun06:
This is a great addition. Can this we added soon?

ianyh:
This is pretty cool. I think we should try to get this merged. I'll do a review when I can.

Amethyst/Layout/ReflowOperation.swift
```diff
@@ -62,6 +62,177 @@ open class ReflowOperation: Operation {
         }
     }
 
+    open func isDegenerateWindow(window: SIWindow) -> Bool {
```
ianyh: What cases did you encounter for which this actually returned `true`?

Amethyst/Layout/ReflowOperation.swift
```diff
@@ -62,6 +62,177 @@ open class ReflowOperation: Operation {
...
+                    let frame_center_x = frame.origin.x + (frame.width/2.0)
+                    let frame_center_y = frame.origin.y + (frame.height/2.0)
+                    let distance_x = (frame_center_x - windows_center_x)
+                    let distance_y = (frame_center_y - windows_center_y)
+                    let distance = sqrt(distance_x*distance_x + distance_y*distance_y)
+                    if Float(distance) < min_distance {
+                        min_distance = Float(distance)
+                    }
+                }
+                if min_distance < 1.0 {
```
ianyh: Some windows don't align perfectly. Terminal windows, for example, are quantized to the size of the cursor. Doesn't this mean such windows will be incorrectly categorized as displaced?

Amethyst/Layout/ReflowOperation.swift
```diff
@@ -62,6 +62,177 @@ open class ReflowOperation: Operation {
...
+        //Now preferentially assign displaced windows a location
+        let screenFrame = adjustedFrameForLayout(screen)
+        let focusedWindow = SIWindow.focused()
+        var frameAssignments: [FrameAssignment] = []
+        for window in displacedWindows {
+            var min_distance = FLT_MAX
+            var min_index = -1
+
+            //Sometimes the windows become degenerate for some reason
+            if !isDegenerateWindow(window:window) {
```
ianyh: This degenerate window thing is worrying me. It feels like there's a deeper problem there that could be the source of existing bugs elsewhere.

djp3:
I'll see if I can bring it up to date with the current architecture.

djp3:
I love this idea, but the code base has moved on and the current code isn't going to ever merge.

PR #536, opened 2017-01-13
Animate window movement

djp3:
This is a feature that adds a rough animation to the windows as they are repositioned.  This helps to keep track of where your windows are so that you don't waste time reorienting yourself after a window relayout event.

There is a check box in the preference pane which turns it on and off.

It animates whenever a reflow event occurs, so if a new window pops up, a window goes away, or a layout style shift is initiated.    

See attached video with demo.
[Terminal_ScreenShot_002.mp4.zip](https://github.com/ianyh/Amethyst/files/704492/Terminal_ScreenShot_002.mp4.zip)

I'm into it.


djp3:
Okay, I accidentally merged multiple forks. argh.

PR #537, opened 2017-01-13
Animate windows clean branch

djp3:
This is a feature that adds a rough animation to the windows as they are repositioned.  This helps to keep track of where your windows are so that you don't waste time reorienting yourself after a window relayout event.

There is a check box in the preference pane which turns it on and off.

It animates whenever a reflow event occurs, so if a new window pops up, a window goes away, or a layout style shift is initiated.    

See attached video with demo.
[Terminal_ScreenShot_002.mp4.zip](https://github.com/ianyh/Amethyst/files/704492/Terminal_ScreenShot_002.mp4.zip)

I'm into it.

[Trello Card](https://trello.com/c/ZuLGfnYU/253-amethyst-animate-windows-clean-branch)

### DIFF
--- a/Amethyst/Events/HotKeyManager.swift
+++ b/Amethyst/Events/HotKeyManager.swift
@@ -188,6 +188,10 @@ open class HotKeyManager: NSObject {
         constructCommandWithCommandKey(CommandKey.ToggleFocusFollowsMouse.rawValue) {
             self.userConfiguration.toggleFocusFollowsMouse()
         }
+        
+        constructCommandWithCommandKey(CommandKey.ToggleAnimateWindows.rawValue) {
+            self.userConfiguration.toggleAnimateWindows()
+        }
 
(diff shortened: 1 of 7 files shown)


ianyh:
Hm. This is kind of cool, but I have two concerns.

One is how expensive the transitions become. The current ones can already be kind of expensive under load, resulting in unpleasant delays, but the changes still get queued up so it can all come crashing down after a bit, even if you're currently doing something with the windows. Even in that video the transitions seemed a little choppy and that was just opening some terminals. Also, with larger screens the number of intermediate steps would get large really fast.

The other is that future improvements might break this. For example, one of the technical things that I think could help with some of the problems I mentioned above is dropping transitions that have been superseded by future transitions, but that is fundamentally in opposition to how this works. The intermediate frames would almost all get dropped.

ianyh:
Closing due to inactivity.

Issue #73, opened 2013-10-30
Maybe add some info about disabling automatically rearranging spaces


kozikow:
Default behaviour of mac spaces is quite horrible - mac rearranged spaces based on recent use. I can't have certain space for certain activities, like space 1 for web, space 2 for coding, space 3 for chat etc. 
I would add section to readme about disabling this behaviour: http://www.macobserver.com/tmo/answers/prevent_lions_spaces_from_automatically_changing_order . 

[Trello Card](https://trello.com/c/b9w1gASO/27-amethyst-maybe-add-some-info-about-disabling-automatically-rearranging-spaces)

ianyh:
Closed by the above PR.

benwinding:
Love the project! but this issue was bugging me for a while before I found the issue:

![image](https://user-images.githubusercontent.com/11782590/127423014-1faa4a7f-9225-42ea-b400-56b994e2796f.png)

This information was originally merged into the README but seems to of been removed again by this commit a3b6a8e46e035220cd09771b92a9b7a643bdd0d5

Since I found this issue to be super annoying I've made a small PR to add it back in!

PR #1138

PR #557, opened 2017-03-01
Update README to include mention of Spaces auto-rearrangement

zackham:
I imagine this has bit more people than just myself, but I thought I was going crazy with my Spaces periodically not being where I expected them to be when I hit the `ctrl-[n]` hotkeys. I searched issues, read the README, and finally ended up just googling around before I found out that this is just a setting in Mission Control.  Seemed like it was worth explicitly calling out in the README.


ianyh:
This closes #73 

Thanks!

PR #580, opened 2017-04-24
Add main pane adjustments for row and column layouts

ianyh:


ianyh:
@glibsm FYI, I fixed up #497 for recent code changes and implemented the changes for column layout as well.

PR #595, opened 2017-05-11
Adds hotkey support for 10th space

### LINKED ISSUE #594: Support 10th space in mod2 + [n]?
Hi, love using Amethyst! I was curious about if there were a specific reason why the `mod2 + [n]` shortcut doesn't support a 10th space where macOS supports it via `ctrl + 0`?

Perhaps I'm odd but I like having a specific application in the last space in my ten space setup.

Looks like the code change required to support that is very straight-forward so I could submit a PR if there isn't something I'm missing for it to be done in the first place.

Thanks!

dennistang:
Closes #594.

Not the prettiest solution, as it mucks up the simple way of setting up the hotkeys from spaces 1-9. Open to suggestions on how to better map the `0` key to send `throw-space-10`.

### Observations
* Couldn't keep the loop simple using `0..<10` since it messes up the order of hotkey in the app's preferences, or without messing with `WindowModifier`. Figured it better to remap in `HotKeyManager` if we had to do it at all.

### DIFF
--- a/Amethyst/Events/HotKeyManager.swift
+++ b/Amethyst/Events/HotKeyManager.swift
@@ -154,7 +154,7 @@ final class HotKeyManager: NSObject {
             }
         }
 
-        (1..<10).forEach { spaceNumber in
+        (1...10).forEach { spaceNumber in
             let commandKey = "\(CommandKey.throwSpacePrefix.rawValue)-\(spaceNumber)"
 
             self.constructCommandWithCommandKey(commandKey) {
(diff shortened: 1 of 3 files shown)


Amethyst/Events/HotKeyManager.swift
```diff
@@ -323,9 +324,15 @@ final class HotKeyManager: NSObject {
         hotKeyNameToDefaultsKey.append(["Swap focused window with main window", CommandKey.swapMain.rawValue])
         hotKeyNameToDefaultsKey.append(["Force windows to be reevaluated", CommandKey.reevaluateWindows.rawValue])
 
-        (1..<10).forEach { spaceNumber in
+        (1...10).forEach { spaceNumber in
             let name = "Throw focused window to space \(spaceNumber)"
-            hotKeyNameToDefaultsKey.append([name, "\(CommandKey.throwSpacePrefix.rawValue)-\(spaceNumber)"])
+
+            // Need to map spaceNumber 10 to throw-space-0 to preserve order of hotkeys
```
ianyh: What order isn't maintained?

dennistang: Doing `(0..<10)` would affect the order of hotkeys in the application preferences:
bildschirmfoto 2017-05-11 um 17 37 50

ianyh: Is there a reason just letting it be 10 instead of mapping it to 0 for the strings doesn't work? It seems like the only thing that actually needs to be 0 is the command's key in `default.amethyst`.

dennistang: You're totally right, not sure why I overcomplicated it. I'll re-submit the PR.

ianyh: No worries. Thanks for the change!

PR #596, opened 2017-05-13
Fix dead Trans*H4CK link.

mclehman:
Wasn't sure whether the homepage or donation page would be better, but I figure the donation page is more similar to the support page linked prior.


ianyh:
Oh thanks! Yeah the donation page is ideal.

PR #621, opened 2017-07-14
Always swap windows if there are only 2 of them

### LINKED ISSUE #620: If there are only 2 windows, "Swap focused window with main window" should work no matter which is focused
## System

 - **Amethyst version**: 0.11.2 (45)
 - **macOS version**: 10.12.5

## What's the feature request?

In the normal case, the "Swap focused window with main window" command doesn't do anything if the main window is focused.  Presumably this is because several non-main windows might exist, and it wouldn't be clear which one of them needed to be swapped with the main window.

But when there are only 2 windows, the swap operation should work regardless of which one is focused.  This is especially important since the "main" window may not be the largest window when there are only 2 windows.

ifreecarve:
* Fixes #620

### DIFF
--- a/Amethyst/Managers/WindowModifier.swift
+++ b/Amethyst/Managers/WindowModifier.swift
@@ -116,11 +116,19 @@ extension SingleScreenWindowMover where Self: ScreenFocuser {
 
         let windows = activeWindows(on: screen)
 
-        guard windows.index(of: focusedWindow) != nil else {
+        guard let focusedIndex = windows.index(of: focusedWindow) else {
             return
         }
 
-        switchWindow(focusedWindow, with: windows[0])
+        // if there are 2 windows, we can always swap.  Just make sure we don't swap focusedWindow with itself.
+        switch windows.count {
+        case 1:
+            return
+        case 2:
+            switchWindow(focusedWindow, with: windows[1 - focusedIndex])
+        default:
+            switchWindow(focusedWindow, with: windows[0])
+        }
         focusedWindow.am_focusWindow()
     }
 


bin/setup.sh
```diff
@@ -1,9 +1,8 @@
 #!/usr/bin/env bash
 
-# Check for Carthage
-printf "Checking for Carthage: "
-carthage version >/dev/null 2>&1 || { printf >&2 "nope!\nCarthage must be installed. Try running:\n\tbrew install carthage\n"; exit 1; }
-printf "found!\n"
+# Check Xcode version (TODO: semver)
+printf "Checking Xcode version: "
+[ `xcodebuild -version | grep Xcode | awk '{print $2}'` = "9.0" ] || { printf >&2 "nope!\nXcode 9.0 must be installed\n"; exit 1; }
```
ianyh: This, unfortunately, doesn't work for two reasons:
 - It only reports the command-line tools version for the selected instance, of which there are many possibilities, so this could fail for someone who has Xcode 8.3 and Xcode 9 installed simultaneously, and
 - it doesn't take into account future versions of Xcode.

As for the first problem, this could probably be downgraded to an informative warning rather than an error.

I'm less sure how to fix the second problem. String ordering would work for versions until we hit Xcode 10. My bash-fu is not strong in this area. Any ideas?

bin/setup.sh
```diff
@@ -12,11 +11,23 @@ printf "found!\n"
...
+    printf "Ensuring correct ruby version is installed:\n"
+    rbenv install -s
+    printf "Correct ruby version installed!\n"
+
+else
+    printf "nope!\n"
+    ACTUAL=`ruby -v | awk '{print $2}' | awk -Fp '{print $1}' | tr -d '\n'`
+    EXPECTED=`cat .ruby-version | tr -d '\n'`
+    if [ $ACTUAL = $EXPECTED ] ; then
+	printf "Correct ruby version is already installed, without rbenv\n"
```
ianyh: I'm honestly a little wary about not blocking on `rbenv`. The nice thing about it is it pretty much removes a slew of debugging questions that aren't even really relevant to the codebase. It lets us not care about global ruby version changes or local ruby version requirement changes; system ruby usually requires root privileges to install gems; and so on. Do you have thoughts on why it should be a looser requirement?

ianyh:
@ifreecarve Actually, could you open separate PRs? I don't want your window swap change to get lost in discussion of the setup script.

ifreecarve:
I've removed the shell script stuff from this PR

ianyh:
Great! Thanks.

PR #624, opened 2017-07-16
Update setup script

ifreecarve:
* Carthage isn't used, so don't check for it
* Xcode 9 is required, so print a warning if it's not `xcode-select`ed
* Relax the check on `rbenv` to just a warning if it doesn't exist

I relaxed the check on `rbenv` because it works on my machine without installing `rbenv`.  It looks like the only real purpose of rbenv is to install bundler, and the only real purpose of bundler is to install Cocoapods.  So if the correct version of Cocoapods is available, then the other steps are (in my opinion) somewhat optional.

### DIFF
--- a/bin/setup.sh
+++ b/bin/setup.sh
@@ -30,7 +30,7 @@ else
     if [ $ACTUAL = $EXPECTED ] ; then
 	printf "Correct ruby version is already installed, without rbenv\n"
     else
-	printf "WARNING: You have ruby $ACTUAL, we want ruby $EXPECTED, and and rbenv is not installed.\n"
+	printf "WARNING: You have ruby $ACTUAL, we want ruby $EXPECTED, and rbenv is not installed.\n"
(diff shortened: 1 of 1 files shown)


bin/setup.sh
```diff
@@ -12,11 +17,23 @@ printf "found!\n"
...
+    printf "Correct ruby version installed!\n"
+
+else
+    printf "nope!\n"
+    ACTUAL=`ruby -v | awk '{print $2}' | awk -Fp '{print $1}' | tr -d '\n'`
+    EXPECTED=`cat .ruby-version | tr -d '\n'`
+    if [ $ACTUAL = $EXPECTED ] ; then
+	printf "Correct ruby version is already installed, without rbenv\n"
+    else
+	printf "WARNING: You have ruby $ACTUAL, we want ruby $EXPECTED, and and rbenv is not installed.\n"
```
ianyh: Remove second "and."

ianyh:
Looks good! Thanks for the contributions!

Issue #9, opened 2013-06-10
Dragging a window with the mouse should correctly move it to whatever managed screen/space it was moved to


ianyh:
If I drag a window to another screen, it should then tile on that screen with the appropriate tiling algorithm. Otherwise it just snaps back.

[Trello Card](https://trello.com/c/1UyUlJod/17-amethyst-dragging-a-window-with-the-mouse-should-correctly-move-it-to-whatever-managed-screen-space-it-was-moved-to)

ianyh:
I think fixed by #627

PR #627, opened 2017-07-19
Move and Resize Windows Using The Mouse

### LINKED ISSUE #216: iTerm constantly resizing
I have a second external monitor (24", 1920x1280) in vertical mode. I position 2 iterm windows there. The layout is 'Rows'.

What happens now is that the windows constantly resize. If i give focus to the bottom one, it increases its height and overlaps with the top window and vice versa.

The "solution" is to put a third window there temporarily (Chrome in my case) and after the resizing is done, throw the third window to the other screen. Now both iterm windows are at exactly have the size and don't consistently try to resize themselves.

I am running Amethyst 0.8.6.

[Trello Card](https://trello.com/c/6WJ5evfY/73-amethyst-iterm-constantly-resizing)

### LINKED ISSUE #341: Outlook 2011 doesn't resize properly and interferes with other apps on the same screen, while using Amethyst
First of all, thank you very much for Amethyst!  I finally get a tiling window manager I can use, for Mac!

I have Yosemite and Amethyst 0.9.6 (27), plus Outlook 2011.

Outlook generally takes the left half of the screen and messes up the right pane, so that other apps resize improperly.

I do have a 3 screen setup (from left to right: Samsung HDMI 1920x1080 on HDMI, Dell Ultrasharp 34 on Thunderbolt -> DisplayPort, and Late 2013 Mac Book Pro w/ Retina), so as a workaround, I move Outlook to my laptop monitor, by itself.

If Outlook is on its own monitor, it still wants to be on the left half of the screen, but at least it doesn't mess up other apps.

Thanks

[Trello Card](https://trello.com/c/EyIIy2ns/169-amethyst-outlook-2011-doesn-t-resize-properly-and-interferes-with-other-apps-on-the-same-screen-while-using-amethyst)

### LINKED ISSUE #617: Resize panes when a window is resized using the mouse
## System

 - **Amethyst version**: 0.11.2
 - **macOS version**: 10.12.5

## What's the Feature Request?
There are keyboard shortcuts to expand or shrink the main pane.  Intuitively, I'd expect that manually resizing the window in the main pane would resize the pane.  The same should happen if I resize the other windows aside from the window in the main pane.  

The difficulty would be in determining what a particular kind of resize means.  Different window resize actions would mean different thing depending on layout, or might be ignored.  Amethyst might need to pop an alert or dialog to allow the user to approve the pane resize.  

[Trello Card](https://trello.com/c/nVSXDGxH/276-amethyst-resize-panes-when-a-window-is-resized-using-the-mouse)

### LINKED ISSUE #618: Mouse drag-drop window to different panes
## System

 - **Amethyst version**: 0.11.2
 - **macOS version**: 10.12.5

## What's the Feature Request?
When I drag a window over a window in another pane, I want Amethyst to swap the windows between those two panes.  We can move windows to different panes (forward and back) by keyboard.  Now the mouse could be used to do the same thing.

[Trello Card](https://trello.com/c/VfmWeHIY/277-amethyst-mouse-drag-drop-window-to-different-panes)

### LINKED ISSUE #634: Amethyst should only move windows as part of a reflow operation -- not on focus changes
I'd like to refactor the discussion from #216 (on handling of windows that don't obey Accessibility API sizing) into this issue, possibly closing the old one.

## System

 - **Amethyst version**: 0.11.3
 - **macOS version**: 10.12.5

## What's the problem?

Changing the focus of windows sometimes causes `kAXWindowMovedNotification` events, because those windows really are being moved.  

 
> Scenarios like this are pretty common with terminal windows. 

> [these windows] don't actually get sized to the exact specifications provided via the Accessibility API; [their] window sizes are based on cursor size 

> Consider the case where we want the window to be (200, 200) and the window sizes itself to (200, 210). Then we add another window and we want it to be (200, 100) and the window sizes itself to (200, 95). Then we try to size it back up and it sizes it to (200, 195). 

> [this] triggers a usually useful behavior: windows that are too big to fit into their space in the grid are pushed on and off screen according to their focus.

## How can it be reproduced?

1. Open 3 Terminal windows in a `Tall` window layout
2. Adjust the size of the main window until the smaller terminal windows are just larger than their frame
3. Change focus between the terminal windows. 

Notice that the terminal windows are pushed slightly off-screen when they lose focus, and pushed on-screen when they gain focus.  `kAXWindowMovedNotification` events are triggered as a result.

## Desired Fix

> Sizing up and sizing down seem to have different effects. When the window is smaller than desired you just end up with a bit of empty space instead of overlapping windows.

Where possible (and if possible) we should attempt to force windows to a size slightly smaller than the frame. This would eliminate the need to move windows on and off screen with focus changes.  This might require some sort of small binary search with respect to window size, to find the most appropriate dimensions for the available frame.

If the window refuses to be shrunken into a given frame, the window should simply be allowed to flow outside that frame (to the right, and to the bottom, specifically).  The window's stubbornness is ultimately the user's problem -- not Amethyst's.  The user should use the layout options available to them to put the window in a location that makes sense (i.e. floating, fullscreen, or just in the upper left of any given layout).  

In all cases, Amethyst should not be moving windows outside of a reflow operation.

[Trello Card](https://trello.com/c/Uxr4Znbd/282-amethyst-amethyst-should-only-move-windows-as-part-of-a-reflow-operation-not-on-focus-changes)

### LINKED ISSUE #646: Secondary windows are drawn on top of > 1 main windows in row and column layouts
## System

 - **Amethyst version**: 0.11.3
 - **macOS version**: 10.12.5

## What's the problem?
When placing secondary windows in row and column layouts, their offset is not calculated properly and they will overlap all but the first "main" window.

## How can it be reproduced?
1. open 3 terminal windows
2. select column layout
3. increment number of main windows

the secondary window now overlaps the second main window

any

no

[Trello Card](https://trello.com/c/OcmpqX9y/287-amethyst-secondary-windows-are-drawn-on-top-of-1-main-windows-in-row-and-column-layouts)

ifreecarve:
## Demo

amethyst_mouse_demo_480

* Fixes #618 
* Fixes #634 
* Fixes #646
* Fixes #617 
* Fixes #216  (More accurately, can't reproduce it after making possibly-relevant changes; assuming fixed)
* Fixes #341  (More accurately, can't reproduce it after making possibly-relevant changes; assuming fixed)

## Current issues

* I do not aggressively reposition windows that are dragged to a different location within their same frame.  I'm not sure that I should.

## Features That Look Like Bugs

* When you resize a window, you are resizing _that window_ -- not a pane --, even if the border of that window lies on a pane boundary.  This means that for column layout, or multiple main windows, your newly-resized window may be nudged into a different location from the one where you released the mouse button; this happens when layouts constrain many windows to be the same size.  The simplest example of this would be the middle layout when 3 windows are open; the window will recenter after you resize, giving the appearance that you're only effecting half of the drag operation.
* Windows that enforce their own minimum size may overlap with other windows that attempt to resize into that frame.  If you want to arrange your windows such that the window with a size constraint is at its minimum (or maximum) size, then that's the window you should be resizing.

## Testing

* Tested on all layouts.
* Resize logic tested on all layouts, with and without multiple main windows
* Tested with focus-follows-mouse enabled
* Tested with window margins on 
* Tested with multiple monitors: 
  * a combination of a vertical and a horizontal monitor
  * in a configuration where the global coordinate space is larger than either of the individual displays.  
* Tested alongside "swap focused window with main window" actions

### Handling the mouse
Subscribe to two event streams: the mouse (`.leftMouseUp` events), and accessibility (`kAXWindowMovedNotification` / `kAXWindowResizedNotification` events).  Get the window being moved from `kAXWindowMovedNotification`, and the moment that the drag ends from `.leftMouseUp`.   We use a small delay hack here to prevent race conditions, since these 2 streams are in no way synchronized.

### Swapping windows
When `kAXWindowMovedNotification` and the mouse button is down, capture a reference to that window. When the mouseUp event fires, read the mouse pointer position.  Find the window underneath the one we are dragging (using code refactored from the "swap with focused window" algorithm), and swap it with whatever we are dragging.  Reflow all windows except the one we are dragging.

This approach is somewhat sane because due to the tiling we are reasonably guaranteed to avoid the situation where an external window-move operation happens with 2 windows simultaneously under the mouse pointer.

The feature itself is behind a checkbox in the preferences window.

### Resizing windows
When `kAXWindowResizedNotification` if the mouse is down, capture the layout and find how big the window is supposed to be.  Then calculate how big the pane would have to be to reflow the window to its current size, and save that information for later.  When the event fires, apply the change.

This approach is somewhat sane because since we're following the mouse button we're unlikely to end up in some sort of reflow/resize loop.

The feature itself is behind a checkbox in the preferences window.

screen shot 2017-08-24 at 1 48 21 pm

[Trello Card](https://trello.com/c/EvSDKLAD/280-amethyst-swap-windows-between-frames-using-the-mouse)

### DIFF
--- a/Amethyst/Categories/SIWindow+Amethyst.swift
+++ b/Amethyst/Categories/SIWindow+Amethyst.swift
@@ -45,12 +45,12 @@ extension SIWindow {
     }
 
     static func topWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow]) -> SIWindow? {
-        let info = windowInformation(windows)
-        guard let windowDescriptions = info.descriptions, !windowDescriptions.isEmpty else {
+        let (ids, maybeWindowDescriptions) = windowInformation(windows)
+        guard let windowDescriptions = maybeWindowDescriptions, !windowDescriptions.isEmpty else {
             return nil
         }
 
-        let windowsAtPoint = onScreenWindowsAtPoint(point, withIDs: info.IDs, withDescriptions: windowDescriptions)
+        let windowsAtPoint = onScreenWindowsAtPoint(point, withIDs: ids, withDescriptions: windowDescriptions)
 
         guard !windowsAtPoint.isEmpty else {
             return nil
@@ -86,20 +86,18 @@ extension SIWindow {
 
     static func alternateWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow], butNot ignoreWindow: SIWindow?) -> SIWindow? {
         // only consider windows on this screen
-        let info = windowInformation(windows)
-        guard let windowDescriptions = info.descriptions, !windowDescriptions.isEmpty else {
+        let (ids, maybeWindowDescriptions) = windowInformation(windows)
+        guard let windowDescriptions = maybeWindowDescriptions, !windowDescriptions.isEmpty else {
             return nil
         }
 
-        let windowsAtPoint = onScreenWindowsAtPoint(point, withIDs: info.IDs, withDescriptions: windowDescriptions)
+        let windowsAtPoint = onScreenWindowsAtPoint(point, withIDs: ids, withDescriptions: windowDescriptions)
 
         for windowDescription in windowsAtPoint {
             if let window = windowInWindows(windows, withCGWindowDescription: windowDescription) {
                 if window != ignoreWindow {
                     return window
                 }
-            } else {
-                print("Couldn't get window \(windowDescription["kCGWindowName"])")
             }
         }
 
--- a/Amethyst/Categories/SIWindow+Amethyst.swift
+++ b/Amethyst/Categories/SIWindow+Amethyst.swift
@@ -46,13 +46,13 @@ extension SIWindow {
 
     static func topWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow]) -> SIWindow? {
         let info = windowInformation(windows)
-        guard let windowDescriptions = info.descriptions, windowDescriptions.count > 0 else {
+        guard let windowDescriptions = info.descriptions, !windowDescriptions.isEmpty else {
             return nil
         }
 
         let windowsAtPoint = onScreenWindowsAtPoint(point, withIDs: info.IDs, withDescriptions: windowDescriptions)
 
-        guard windowsAtPoint.count > 0 else {
+        guard !windowsAtPoint.isEmpty else {
             return nil
         }
 
@@ -87,7 +87,7 @@ extension SIWindow {
     static func alternateWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow], butNot ignoreWindow: SIWindow?) -> SIWindow? {
         // only consider windows on this screen
         let info = windowInformation(windows)
-        guard let windowDescriptions = info.descriptions, windowDescriptions.count > 0 else {
+        guard let windowDescriptions = info.descriptions, !windowDescriptions.isEmpty else {
             return nil
         }
 
--- a/Amethyst/Categories/SIWindow+Amethyst.swift
+++ b/Amethyst/Categories/SIWindow+Amethyst.swift
@@ -11,30 +11,46 @@ import Foundation
 import Silica
 
 extension SIWindow {
-    static func topWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow]) -> SIWindow? {
-        guard let windowDescriptions = windowDescriptions(.optionOnScreenOnly, windowID: CGWindowID(0)), windowDescriptions.count > 0 else {
-            return nil
-        }
+    fileprivate static func windowInformation(_ windows: [SIWindow]) -> (IDs: Set<CGWindowID>, descriptions: [[String: AnyObject]]?) {
+        let ids = Set(windows.map { $0.windowID() })
+        return (IDs: ids, descriptions: windowDescriptions(.optionOnScreenOnly, windowID: CGWindowID(0)))
+    }
 
-        let windowIDs = Set(windows.map { $0.windowID() })
-        var windowsAtPoint: [[String: AnyObject]] = []
+    fileprivate static func onScreenWindowsAtPoint(_ point: CGPoint,
+                                                   withIDs windowIDs: Set<CGWindowID>,
+                                                   withDescriptions windowDescriptions: [[String: AnyObject]]) -> [[String: AnyObject]] {
+        var ret: [[String: AnyObject]] = []
+
+        // build a list of windows at this point
         for windowDescription in windowDescriptions {
-            guard let windowID = (windowDescription[kCGWindowNumber as String] as? NSNumber).flatMap({ CGWindowID($0.intValue) }), windowIDs.contains(windowID) else {
+            guard let windowID = (windowDescription[kCGWindowNumber as String] as? NSNumber).flatMap({ CGWindowID($0.intValue) }),
+                windowIDs.contains(windowID) else {
                 continue
             }
 
+            // only consider windows with bounds
             guard let windowFrameDictionary = windowDescription[kCGWindowBounds as String] as? [String: Any] else {
                 continue
             }
 
+            // only consider window bounds that contain the given point
             let windowFrame = CGRect(dictionaryRepresentation: windowFrameDictionary as CFDictionary)!
-
             guard windowFrame.contains(point) else {
                 continue
             }
-
-            windowsAtPoint.append(windowDescription)
+            ret.append(windowDescription)
         }
+
+        return ret
+    }
+
+    static func topWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow]) -> SIWindow? {
+        let info = windowInformation(windows)
+        guard let windowDescriptions = info.descriptions, windowDescriptions.count > 0 else {
+            return nil
+        }
+
+        let windowsAtPoint = onScreenWindowsAtPoint(point, withIDs: info.IDs, withDescriptions: windowDescriptions)
 
         guard windowsAtPoint.count > 0 else {
             return nil
@@ -66,6 +82,27 @@ extension SIWindow {
         }
 
         return windowInWindows(windows, withCGWindowDescription: windowDictionaryToFocus)
+    }
+
+    static func secondWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow]) -> SIWindow? {
+        // only consider windows on this screen
+        let info = windowInformation(windows)
+        guard let windowDescriptions = info.descriptions, windowDescriptions.count > 0 else {
+            return nil
+        }
+
+        let windowsAtPoint = onScreenWindowsAtPoint(point, withIDs: info.IDs, withDescriptions: windowDescriptions)
+
+        // early exit for no 2 windows found
+        guard windowsAtPoint.count > 1 else {
+            return nil
+        }
+
+        // it's the one under the one we're dragging
+        guard let ret = windowInWindows(windows, withCGWindowDescription: windowsAtPoint[1]) else {
+            return nil
+        }
+        return ret
     }
 
     static func windowInWindows(_ windows: [SIWindow], withCGWindowDescription windowDescription: [String: AnyObject]) -> SIWindow? {
--- a/Amethyst/Categories/SIWindow+Amethyst.swift
+++ b/Amethyst/Categories/SIWindow+Amethyst.swift
@@ -84,7 +84,7 @@ extension SIWindow {
         return windowInWindows(windows, withCGWindowDescription: windowDictionaryToFocus)
     }
 
-    static func secondWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow]) -> SIWindow? {
+    static func alternateWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow], butNot ignoreWindow: SIWindow?) -> SIWindow? {
         // only consider windows on this screen
         let info = windowInformation(windows)
         guard let windowDescriptions = info.descriptions, windowDescriptions.count > 0 else {
@@ -93,16 +93,17 @@ extension SIWindow {
 
         let windowsAtPoint = onScreenWindowsAtPoint(point, withIDs: info.IDs, withDescriptions: windowDescriptions)
 
-        // early exit for no 2 windows found
-        guard windowsAtPoint.count > 1 else {
-            return nil
+        for windowDescription in windowsAtPoint {
+            if let window = windowInWindows(windows, withCGWindowDescription: windowDescription) {
+                if window != ignoreWindow {
+                    return window
+                }
+            } else {
+                print("Couldn't get window \(windowDescription["kCGWindowName"])")
+            }
         }
 
-        // it's the one under the one we're dragging
-        guard let ret = windowInWindows(windows, withCGWindowDescription: windowsAtPoint[1]) else {
-            return nil
-        }
-        return ret
+        return nil
     }
 
     static func windowInWindows(_ windows: [SIWindow], withCGWindowDescription windowDescription: [String: AnyObject]) -> SIWindow? {
--- a/Amethyst/Categories/SIWindow+Amethyst.swift
+++ b/Amethyst/Categories/SIWindow+Amethyst.swift
@@ -11,6 +11,8 @@ import Foundation
 import Silica
 
 extension SIWindow {
+    // convert SIWindow objects to CGWindowIDs.
+    // additionally, return the full set of window descriptions (which is unsorted and may contain extra windows)
     fileprivate static func windowInformation(_ windows: [SIWindow]) -> (IDs: Set<CGWindowID>, descriptions: [[String: AnyObject]]?) {
         let ids = Set(windows.map { $0.windowID() })
         return (IDs: ids, descriptions: windowDescriptions(.optionOnScreenOnly, windowID: CGWindowID(0)))
@@ -44,6 +46,7 @@ extension SIWindow {
         return ret
     }
 
+    // if there are several windows at a given screen point, take the top one
     static func topWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow]) -> SIWindow? {
         let (ids, maybeWindowDescriptions) = windowInformation(windows)
         guard let windowDescriptions = maybeWindowDescriptions, !windowDescriptions.isEmpty else {
@@ -84,6 +87,7 @@ extension SIWindow {
         return windowInWindows(windows, withCGWindowDescription: windowDictionaryToFocus)
     }
 
+    // get the first window at a certain point, excluding one specific window from consideration
     static func alternateWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow], butNot ignoreWindow: SIWindow?) -> SIWindow? {
         // only consider windows on this screen
         let (ids, maybeWindowDescriptions) = windowInformation(windows)
@@ -104,6 +108,7 @@ extension SIWindow {
         return nil
     }
 
+    // find a window based on its window description within an array of SIWindow objects
     static func windowInWindows(_ windows: [SIWindow], withCGWindowDescription windowDescription: [String: AnyObject]) -> SIWindow? {
         for window in windows {
             guard
@@ -132,6 +137,8 @@ extension SIWindow {
         return nil
     }
 
+    // return an array of dictionaries of window information for all windows relative to windowID
+    // if windowID is 0, this will return all window information
     static func windowDescriptions(_ options: CGWindowListOption, windowID: CGWindowID) -> [[String: AnyObject]]? {
         guard let cfWindowDescriptions = CGWindowListCopyWindowInfo(options, windowID) else {
             return nil
--- a/Amethyst/Layout/Layout.swift
+++ b/Amethyst/Layout/Layout.swift
@@ -32,22 +32,7 @@ struct FrameAssignment {
             finalFrame.size.height -= 2 * padding
         }
 
-        var finalPosition = finalFrame.origin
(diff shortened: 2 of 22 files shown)


ifreecarve:
This has some handy debug statements that you can watch while you drag.

[amethyst_mouse_debugging.patch.txt](https://github.com/ianyh/Amethyst/files/1158898/amethyst_mouse_debugging.patch.txt)

ianyh:
I'm still concerned about it automatically switching windows while in-flight. The problem is that they don't strictly move out of the way, they get swapped. For windows that are adjacent in the window list this isn't an issue, but when windows are adjacent on screen but not adjacent in the window list the swap is actually destructive to the layout in a way that the user might not intend.

I can see two things the user might expect. Either,
 - the windows are swapped (i.e., they have switched positions in the window list), or
 - the windows are shifted over (i.e., the window has been removed from the list and inserted in a new position)

The former corresponds to the swap with main command, and the latter corresponds with pushing windows (counter-)clockwise. Honestly, I think the latter makes more intuitive sense to me in terms of how drag and drop reordering tends to work, and it has the advantage of being easily reversible--just drag back to where it was--but I could see it going either way for any one person.

ianyh:
Also, does this work with focus-follows-mouse?

ifreecarve:
This does work with focus-follows-mouse.  Or at least, I haven't found a condition where dragging a window will accidentally focus a window that isn't being dragged.

I see what you mean about the in-flight swap behavior causing additional windows to move location.  I've been content with this, because it feels better than no mouse support whatsoever.  

I'm content to trigger the swap behavior using a more sophisticated method of drag detection (i.e. waiting for the mouse button to come up), but I haven't found a good model online for how that might work.  

In the meantime, should this go behind a preferences checkbox?

ifreecarve:
Updates:
* Rudimentary detection of drag completion (by monitoring both mouse and Accessibility events) seems to work, and reliably produces the desired effect of reflows not happening until a drag-to-move-window operation has ended.
* The behavior is behind a preferences toggle

ifreecarve:
Update: Problems based on dragging a floating window (both kinds -- those specified in preferences and those that offer a child "tool window" that's not coerced into a frame) have been fixed.

I have also fixed #634 in preparation for messing with resize events.

ifreecarve:
A month goes by fast!

Fixed 2 bugs in this branch:
* Floating windows (i.e. specified in settings) and non-managed windows (e.g. Giphy Capture) were causing swaps of the windows beneath them when they were dragged.  This was fixed by asking the layout what windows are being managed.
* Sometimes swapping a window (especially terminal windows with Chrome, for some reason) would cause an immediate duplicate swap.  This was fixed by ensuring that only one mouse state was shared between all observers.

Amethyst/Managers/WindowManager.swift
```diff
@@ -77,6 +157,53 @@ private struct ObserveApplicationNotifications {
...
+                guard windowManager.userConfiguration.mouseSwapsWindows() else {
+                    return
+                }
+
+                guard let movedWindow = accessibilityElement as? SIWindow else {
+                    return
+                }
+
+                // Do a more in-depth check; make sure the window being dragged is being managed
+                guard let screenManager = windowManager.focusedScreenManager(),
```
ifreecarve: I may be confused too.  I tried guarding for `window.isActive()` and it never fired.  Is that something totally different?

ifreecarve: Update: just tried this code:

```swift
                guard windowManager.windowIsActive(movedWindow) else {
                    print("BUT WINDOW IS NOT ACTIVE")
                    return
                }

                // Do a more in-depth check; make sure the window being dragged is being managed
```

The "more in-depth check" fires, the print statement does not.

ifreecarve:
To recap a gitter discussion: I believe this branch has now reached MVP status for mouse integration.

ifreecarve:
Resize using mouse has been added to this PR and seems to work well

LanderN:
Resize with mouse doesn't seem to work well with Window Margins enabled (windows overlap after resizing).

ifreecarve:
Hmm, I see this problem on my machine too.  @ianyh will have to weigh in here because, despite very reproducible behavior, I can't seem to explain it.  TL;DR: all windows reflow properly, except for the one you were dragging.  If you use the hotkey (i.e. no window is being dragged), then it works fine.  And even then, there are other effects related to dragging the "free" dimension, the constrained dimension, and both at once..

If I drag the window in the main pane, that window (chrome, in this example) spills over in the +x direction:

screen shot 2017-08-27 at 9 09 35 pm

If I drag-to-resize one of the windows in the secondary pane, that window will also spill over in the +x direction:

screen shot 2017-08-27 at 9 15 20 pm

I would say that this is a result of the mouse-resized window not being reflowed, but that's clearly happening: `layout.expainMainPane` and `layout.shrinkMainPane` are both calling `layout.setMainPaneRatio`, which is how the mouse sets it.  In fact, the resize-by-hotkey and resize-by-mouse use the same code.  I even went as far as to make sure (locally) that the reflow operation is being called (`ScreenManager.updateCurrentLayout`) for both instances.

Long story short, this seems to be related to the reflow operation itself, or to some interaction between the mouse and the reflow operation that I can't explain.

ifreecarve:
@LanderN I think found the problem, and it's unrelated to my patch.  It seems to be a problem in Silica's `SIAccessibilityElement.m` where resize operations are ignored for changes less than 25px.  I'm not sure why using window margins causes a problem there, but commenting out those lines in the Silica file seems to fix the problem:

```objc
- (void)setFrame:(CGRect)frame {
    // We only want to set the size if the size has actually changed.
    BOOL shouldSetSize = self.isResizable;
//    CGRect currentFrame = self.frame;
//    if (self.isResizable) {
//        if (fabs(currentFrame.size.width - frame.size.width) < 25) {
//            if (fabs(currentFrame.size.height - frame.size.height) < 25) {
//                shouldSetSize = NO;
//            }
//        }
//    } else {
//        shouldSetSize = NO;
//    }
```

@ianyh should I submit this as a patch to Silica?

ifreecarve:
Patch submitted to Silica: https://github.com/ianyh/Silica/pull/9

ifreecarve:
## All Systems Go

Happy 3 month anniversary of this PR!  
* I've just pulled in the latest Silica changes
* I've tweaked a small calculation -- resizing windows with padding enabled is now pixel-perfect (except when the padding value is an odd number, which breaks the `floor(UserConfiguration.shared.windowMarginSize() / 2) * 2` calculation -- in which case, the window size moves to a multiple of 2px the first time you resize and is stable after that).
* I've been running my branch out of Xcode for at least the past 8-10 weeks and haven't noticed any strange behavior.

ianyh:
Thanks for the hard work on this. I'll try to take a look soon.

Amethyst/Managers/WindowManager.swift
```diff
@@ -13,6 +13,8 @@ import RxSwiftExt
 import Silica
 import SwiftyJSON
 
+private let mouseDragRaceThresholdSeconds = 0.15
```
ianyh: Global statics like this should be PascalCase not camelCase, but I think this works better inside the `MouseStateKeeper`.
```swift
class MouseStateKeeper {
    private let mouseDragRaceThresholdSeconds = 0.15
    ...
}
```

Amethyst/Managers/WindowManager.swift
```diff
@@ -21,13 +23,111 @@ enum WindowChange {
...
+class MouseStateKeeper {
+    public var state: MouseState
+    private var _windowManager: WindowManager!
+    private var monitor: Any?
+
+    init() {
+        state = .pointing
+    }
+
+    var windowManager: WindowManager {
```
ianyh: I think this is a retain loop with the window manager, and deep linking to the window manager feels a bit odd. Possibly there should be a `MouseStateKeeperDelegate`?

Amethyst/Managers/WindowManager.swift
```diff
@@ -21,13 +23,111 @@ enum WindowChange {
...
+            self._windowManager = newValue
+            let mouseEventsToWatch: NSEvent.EventTypeMask = [.leftMouseDown, .leftMouseUp, .leftMouseDragged]
+
+            // remove any existing monitors, although we expect to never reassign windowManager.
+            if let oldMonitor = monitor {
+                NSEvent.removeMonitor(oldMonitor)
+                monitor = nil
+            }
+
+            monitor = NSEvent.addGlobalMonitorForEvents(matching: mouseEventsToWatch) { anEvent in
```
ianyh: I think this is also a retain loop. The block maintains a strong reference to `self` so it'll never deinit to remove the monitor. In practice this isn't an issue because there only needs to be one for the lifetime of the application, but things might get weird for testing purposes.

Amethyst/Managers/WindowManager.swift
```diff
@@ -21,13 +23,111 @@ enum WindowChange {
...
+                        break // ignore - we have what we need
+                    case .pointing, .clicking, .dragging, .doneDragging:
+                        self.state = .dragging
+                    }
+
+                case .leftMouseUp:
+                    switch self.state {
+                    case .dragging:
+                        // assume window move event will come shortly after
+                        self.state = .doneDragging(atTime: NSDate())
```
ianyh: Use `Date` not `NSDate`.

Amethyst/Managers/WindowManager.swift
```diff
@@ -77,6 +179,79 @@ private struct ObserveApplicationNotifications {
...
+                    return
+                }
+
+                switch self.mouse.state {
+                case .dragging:
+                    // record window and wait for mouse up
+                    self.mouse.state = .moving(window: movedWindow)
+                case let .doneDragging(lmbUpMoment):
+                    // if mouse button recently came up, assume window move is related
+                    let dragEndInterval = NSDate().timeIntervalSince(lmbUpMoment as Date)
```
ianyh: `Date`

Amethyst/Managers/WindowManager.swift
```diff
@@ -77,6 +179,79 @@ private struct ObserveApplicationNotifications {
...
+
+                let ratio = oldFrame.impliedMainPaneRatio(windowFrame: resizedWindow.frame())
+
+                switch self.mouse.state {
+                case .dragging, .resizing:
+                    // record window and wait for mouse up
+                    self.mouse.state = .resizing(screen: screen, ratio: ratio)
+                case let .doneDragging(lmbUpMoment):
+                    // if mouse button recently came up, assume window resize is related
+                    let dragEndInterval = NSDate().timeIntervalSince(lmbUpMoment as Date)
```
ianyh: `Date`

ianyh:
This may fix #9 as well.

Amethyst/Layout/Layout.swift
```diff
@@ -115,15 +131,40 @@ protocol Layout {
     var windowActivityCache: WindowActivityCache { get }
 
     func reflow(_ windows: [SIWindow], on screen: NSScreen) -> ReflowOperation
+    func assignedFrame(_ window: SIWindow, of windows: [SIWindow], on screen: NSScreen) -> FrameAssignment?
 }
 
 protocol PanedLayout {
+    var mainPaneRatio: CGFloat { get }
+    func setMainPaneRawRatio(rawRatio: CGFloat)
```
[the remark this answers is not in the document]
ifreecarve: `shrinkMainPane()` and `expandMainPane()` were already in the protocol, even though their behavior for some layouts is a no-op. So it seemed natural to expose the boilerplate `mainPaneRatio` that was being used to support that feature, even though some layouts ignore it.

There are many ways to do this, and I don't have a preference.  I'm just following the existing implementation, which was in terms of ratios.

Amethyst/Categories/SIWindow+Amethyst.swift
```diff
@@ -11,32 +11,48 @@ import Foundation
 import Silica
 
 extension SIWindow {
-    static func topWindowForScreenAtPoint(_ point: CGPoint, withWindows windows: [SIWindow]) -> SIWindow? {
-        guard let windowDescriptions = windowDescriptions(.optionOnScreenOnly, windowID: CGWindowID(0)), windowDescriptions.count > 0 else {
-            return nil
-        }
+    fileprivate static func windowInformation(_ windows: [SIWindow]) -> (IDs: Set<CGWindowID>, descriptions: [[String: AnyObject]]?) {
+        let ids = Set(windows.map { $0.windowID() })
+        return (IDs: ids, descriptions: windowDescriptions(.optionOnScreenOnly, windowID: CGWindowID(0)))
```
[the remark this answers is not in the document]
ifreecarve: `CGWindowID(0)` was in the original code, likely because it eventually gets passed as `windowID: CGWindowID` to `CGWindowListCopyWindowInfo` (not as an optional).  I'm not sure how to fix this, and the fix itself feels outside the scope of this PR.

ifreecarve:
What's the test case for #9?  I haven't noticed a problem dragging windows between screens (i.e. from a spot on my left monitor to a spot on my right monitor)

Amethyst/Layout/Layout.swift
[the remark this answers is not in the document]
ianyh: The shrink and expand actions are abstract operations. The ratio is an implementation detail.

ianyh:
I don't think you need to test it, I just wanted to make the connection from here to the issue.

Amethyst/Layout/Layout.swift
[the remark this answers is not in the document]
ifreecarve: I'm happy to do it another way if you have something in mind, but (from my reading of the code) the concept of a ratio between a "main" pane and the others seems central to the idea of a `PanedLayout`.  How else would you specify the spatial relationship here?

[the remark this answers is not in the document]
ifreecarve: For context, the problem being solved here is that the `kAXWindowResizedNotification` listener must, without knowing any specifics about the layout, trigger a reflow that will put the window edge exactly where the cursor left it (using nothing more than the resized window's frame information).

The way I'm currently tackling this is to expose `FrameAssignment` and `ResizeRules` for the windows in a layout, which allows me to work backwards from the resized window's frame  and determine what layout parameters would have produced that size.   Then I expose the setting of those parameters via the layout (in this case, the ratio).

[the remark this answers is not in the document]
ianyh: Okay. I think I'm convinced on it being a property of paned layouts in general. It's kind of nit-picking, but I think I would prefer having a method that suggests a ratio rather than just a kind of setter. Maybe
```swift
func reccomendMainPaneRatio(_ ratio: CGFloat)
```
It feels a bit more flexible to me as the layout still maintains the logic of how to actually apply the ratio change. That's just a setter for most (all current?) cases, but could be more complex in the future.

Amethyst/Managers/WindowManager.swift
```diff
@@ -29,33 +27,45 @@ enum MouseState {
     case dragging
     case moving(window: SIWindow)
     case resizing(screen: NSScreen, ratio: CGFloat)
-    case doneDragging(atTime: NSDate)
+    case doneDragging(atTime: Date)
+}
+
+protocol MouseStateKeeperDelegate: class {
+	func focusedScreenManager() -> ScreenManager?
```
ianyh: Minor, but this shouldn't be tab-indented.

Amethyst/Managers/WindowManager.swift
```diff
@@ -29,33 +27,45 @@ enum MouseState {
...
         state = .pointing
     }
 
-    var windowManager: WindowManager {
+    deinit {
+        guard let oldMonitor = monitor else { return }
+        NSEvent.removeMonitor(oldMonitor)
+    }
+
+    var delegate: MouseStateKeeperDelegate {
```
ianyh: Do we still need to gate all of the setup on setting a delegate? It looks like none of the usage of the delegate is synchronously required.

Amethyst/Layout/Layout.swift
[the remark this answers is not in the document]
ifreecarve: Sounds good

Amethyst/Managers/WindowManager.swift
ifreecarve: I agree.  Not sure why Xcode is doing that.

Amethyst/Managers/WindowManager.swift
ifreecarve: Aah, it looks like this logic can be simplified quite a bit.

Amethyst/Categories/SIWindow+Amethyst.swift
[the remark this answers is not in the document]
ifreecarve: Did some more [research](https://stackoverflow.com/a/30337008/2063546) on this.  Apparently the `CGWindowID(0)` is how they do it in CGWindow.h:

```c
/* A value which is never the window ID of any window. */

#define kCGNullWindowID ((CGWindowID)0) 
```

I think the `#define` keeps that constant from being exposed in Swift, so I'm just going to create my own there for clarity.

ianyh:
https://github.com/ianyh/Amethyst/pull/627/commits/b86204c05dd3534fe311025c0f78e156f1513134 is outside of the scope of this change.

ifreecarve:
Removed b86204c from this PR

ifreecarve:
Thanks for your coaching and your high expectations!  If you want to create issues for the things that need moving around, I'd be glad to try and tackle some of them.

Is there a timeline for getting this into a release?  (I'd love to share this with some friends.)

Pitometsu:
What's the common way to install latest version from master? Is there brew recipe?

ianyh:
The version from master is downloadable at https://ianyh.com/amethyst or from the releases page of this project. I think there is a brew recipe for it, and though I sometimes forget about it, it seems to end up getting reasonably maintained.

Getting development versions of the app can only be done through in-app updater or directly downloaded from the releases page of this project. I don't think it can currently be done through homebrew. Unfortunately, this process ends up being largely at the mercy for my schedule, and I am open to suggestions of how to improve it. (Maybe automatic nightly builds?)

fullmetalsheep:
mother of god, i Love this <3

Issue #17, opened 2013-06-18
Should perhaps have a global window minimum


ianyh:
Many windows aren't useful when sized quite small. Things like Xcode recognize this and force a minimum and there's a solution for that (push offscreen when not focused, but move fully onscreen when focused), so maybe we should enforce a global minimum to make sure that windows are actually usefully sized.

[Trello Card](https://trello.com/c/uEA4nIKE/18-amethyst-should-perhaps-have-a-global-window-minimum)

ianyh:
Fixed by #654

Issue #651, opened 2017-10-09
Certain dialog windows are automatically closed by Amethyst


eplanet:
## System
 - **Amethyst version**: 0.11.4
 - **macOS version**: 10.12.6 (16G29)

## What's the problem?
Amethyst seems to be killing a certain type of dialog window. I hope this hasn't been reported / addressed yet, but I couldn't find any related issue.

## How can it be reproduced?
Easily open an existing project structure, then try to add dependencies to it (refer to the screenshot or [official help page](https://www.jetbrains.com/help/idea/working-with-module-dependencies.html#add_dependency)).

## What applications are involved?
* IntelliJ

## Has anything fixed it, even temporarily?
Quitting Amethyst fixes it, that's why I'm reporting the issue here.

## dotfile (if you have one)
None.

## Screenshots
![View of the menu](https://ibb.co/hM4Q1G)

kmajor:
+1 same problem.  It's essentially killing all my chrome dialogs, which means I have to choose between Amethyst and Chrome... :-(

I've tried with both the latest chromium 63.xxxx  and latest Chrome I think 56.xxxx

To reproduce, simply click on any chrome extensions that have a dialog popup and when you mouseover, the window disappears.  

I hope you guys can get this fixed right away, it's probably not widely reported because people can't figure out what's causing it.  I spent hours tracking this down.  

Amethyst is pretty great btw...

ianyh:
This sounds like a problem with focus-follows-mouse. Do you have that option enabled?

eplanet:
That could have explained it but I don't have that option enabled. I tried activating it without success, the same being true for `Mouse follows focused windows` and actually any other option.

spenserw:
I can confirm the problem is with `focus-follows-mouse`. Disabling that option no longer causes the dialogues to close. Same versions as OP.

deiga:
Turning off `focus-follows-mouse` did not work for me. `mouse-follows-focus`could be the culprit for me, but I really need that to stop me from getting Tenosynovitis in my forearm.

axs221:
I think I have a fix for this one. I will get a PR up soon. 

[...]

Issue #495, opened 2016-10-06
Have a configurable minimum size for windows


ianyh:
Right now applications have their own minimum window size and we compensate for that by letting them flow off the edge of the screen when not in focus, and then pulling them on screen when in focus. Someone suggested that it would be nice to be able to configure this behavior manually, e.g., set Terminal to not resize super tiny and instead just flow off screen/get pulled on screen appropriately.

Issue #494, opened 2016-10-06
minimum size for tiled window?


da0ist:
<img width="1440" alt="screen shot 2016-10-06 at 12 26 32 pm" src="https://cloud.githubusercontent.com/assets/7945846/19167608/493f9692-8bc1-11e6-9554-2f2bb7af41bc.png">

Can you make it so apps that actually resize like terminal.app have a minimum size so they pop onto the screen like slack does in this picture?

[Trello Card](https://trello.com/c/lbmRAREy/238-amethyst-minimum-size-for-tiled-window)

ianyh:
Supercedes #495

axs221:
@ianyh I would like to help with this one. It works well if I add a minimum size inside the TallLayout reduce method. I can add a global configuration setting. Here is a video showing it working with it hard-coded to 500px minimum:

https://cl.ly/0H373T2p3f3F

PR #654, opened 2017-10-21
Minimum width and height for active tiled window

axs221:
## Related issue

* #494
* #495
* Possibly affects: #651 (regarding commit 691e27a)

## Motivation

As mentioned in the related issue and its duplicates, it is helpful when 
Amethyst pushes windows off-screen at smaller sizes, and then pulls
those frames back into the window entirely when the window is focused.

However, this behavior was only performed for windows that have sizing
constraints. Additionally, it would be helpful to be able to configure the
minimum size desired when the window is focused and the behavior is
performed.

## Modifications

Configuration settings and related behavior for minimum height and width
of focused window. Minimums only apply to focused window, which helps
minimize times when tiled windows may be hidden from adjacent windows
overlapping if minimum size were applied all the time.

Also minor adjustment to fix how it is determined if frame is within window,
to prevent stuttering issues that would occur if the focus following mouse
option was selected.

## Video

https://cl.ly/2q2h3v0v0Z26

## Screenshots

image

image

[Trello Card](https://trello.com/c/oMOgvetw/291-amethyst-minimum-width-and-height-for-active-tiled-window)


ianyh:
Nice. I'll try to take a more thorough look at this soon, but at first glance it looks good. It looks like the version of CocoaPods using the Gemfile is wrong, though. I'll update that in `development` so you can rebase onto it before merging this.

axs221:
@ianyh Sounds good. I reverted the COCOAPODS version setting in the lock file and have my environment set up with 1.3.1. I left the Gemfile unmodified and I will pull it in from development as you mentioned.

axs221:
I am not sure yet about 691e27a. While it is better for Chrome dialogs on click, it has the side effect of also making all windows of that process focus / zoom when one window is focused. It is mostly an improvement, but need to check how it works on multi-window interfaces like the GIMP, which I am assuming is one process.

axs221:
The "Parent" attribute doesn't get me what I want unfortunately. It gets me process information with a title of the application name and windowID == 0. Comparing parents of Chrome windows is always true for all Chrome windows. I am about out of ideas for how to make dialogs work well without marking all windows of the same process as focused.

ianyh:
The parent of the windows is probably the application element. Sheets can be attached to windows, but aren't necessarily. I don't think modal alerts are attached to anything but the application itself. Regardless, windows of an application are all going to have the same process ID, so I'm not sure the focused check there works.

Preventing window focus when there's a modal for an application seems like a generally good idea, but I would have to think more about what it would look like with multiple applications, each of which might be able to have a dialog open, etc. Either way that should be a separate change from the minimum size thing.

axs221:
@ianyh Sounds good to me. I will revert 691e27a56be8f303a033ca5d05da68da4a50c7dc. It won't provide the best user experience until a solution to have the parent window set as focused, but at least people don't have to use this PR for now if that bothers them.

ianyh:
Should fix https://github.com/ianyh/Amethyst/issues/17

ianyh:
Okay. This looks good, but I'd like to try to get #627 in before merging this. I don't mind handling any merge conflicts that come up.

ianyh:
Looks good! I think I resolved the merge conflicts correctly.

Issue #250, opened 2014-10-20
Add a whitelist of tiling applications


yeban:
I can tell Amethyst which apps to float - https://github.com/ianyh/Amethyst#always-float-an-app. Can it be done the other way round? tell it which ones to tile?

[Trello Card](https://trello.com/c/lE3ySdNC/87-amethyst-add-a-whitelist-of-tiling-applications)

Issue #200, opened 2014-08-25
iTerm2 removing title bar makes amethyst ignore the window


thomaskern:
I compiled iTerm2 after doing this http://www.patrickedelman.com/removing-the-title-bar-from-iterm2/ to it. Now i have no title bar. The thing is, new windows without the title bar are not being resized/placed by amethyst anymore.
why is that?

[Trello Card](https://trello.com/c/8aKYdTmg/68-amethyst-iterm2-removing-title-bar-makes-amethyst-ignore-the-window)

Rnhmjoj:
I didn't compile iterm but in the nightly builds there is now an option to hide the titlebar.
It is causing the same issue. Any updates?

Hultner:
I'm haven't written a line ObjC in probably 7 years and even then barely. Could anyone point me in the right direction for adding a custom patch/hack to fix this behaviour specifically for iTerm/whitelisted applications while waiting for a fix? 
Ping @ianyh @leonardt

neonfrontier:
@Hultner maybe you could take a look at how slate are doing it?
https://github.com/jigish/slate

ddelnano:
Started a fix for it [here](https://github.com/ianyh/Amethyst/pull/374).  The subrole of the title less windows is `AXDialog`. 

The focus aspect needs to be fixed but Amethyst now recognizes that it needs to tile the previously unrecognized windows.

jclusso:
I can confirm this issue too.

ddelnano:
Yea its because applications like Safari and Finder use the `AXDialog` so its not ideal to tile every window that has the `AXDialog` subrole.

0x62:
Still an issue

Ramblurr:
Any updates on this issue? I'd love to use amethyst with a titlebar-less iterm2.

net:
@ianyh status on this issue?

ianyh:
It's tricky because the windows have the subrole `AXDialog`, which is true of a lot of things that you don't want to tile. Popups, for example, are also dialogs. There are probably some ways to differentiate popups from other things (they are modal, they have buttons, etc.). Could also just straight up whitelist iTerm, as there is a ton of demand for this.

ddickstein:
Oh speaking of which, Finder should probably always be treated as a
floating popup. When I open it I generally (1) use it very briefly, and (2)
want to interact with it while it's open (even if in full screen probably
would prefer a small finder appear on top of current app, because usually
[...]

Issue #171, opened 2014-05-02
Option to whitelist tiled windows


glacials:
I treat the `floating` array as a blacklist for programs that don't cooperate well with resizing, or that always have my full attention (like games). It'd be nice to be able to swap this behavior to whitelist style, where I have to add programs to a list in order for them to be tiled.

Thanks!

[Trello Card](https://trello.com/c/dTrr0fVL/53-amethyst-option-to-whitelist-tiled-windows)

kavika13:
@ianyh - I have been using Amethyst for a while now, and I'm finally giving up because I'm spending more time blacklisting and futzing than I am saving at this point. Whitelists IMO are a much better option, as I really only want my dev environment to tile.

kavika13:
@ianyh - I have been using Amethyst for a while now, and I'm finally giving up because I'm spending more time blacklisting and futzing than I am saving at this point. Whitelists IMO are a much better option, as I really only want my dev environment to tile. I think this feature is badly needed.

MaThGo:
Please implement this #200 is a dealbreaker for my workflow.
Otherwise I've to resort to KWM or the unstable ChunkWM.

axs221:
@ianyh I can take a look at this one. I am thinking a checkbox in the Floating section that says "Float all except selected". I haven't looked at where the list is used, but I'm guessing it won't be difficult to invert the check.

axs221:
@strflw I will have a PR up for this soon. As for #200, if you have upgraded to the latest iTerm2 there is a chance iTerm2 will work with Amethyst properly with the built-in iTerm2 "No Title Bar" setting. See #200 for more information, and it would be helpful to know if you can still reproduce at that point.

PR #655, opened 2017-10-22
Use Floating Bundle Identifiers as Blacklist

axs221:
## Related issues

* Main issue: #171
* Duplicate issue: #250 
* An issue where whitelisting was wanted as a workaround: #200 

## Motivation

Some users have mentioned that they either prefer to use Amethyst only for certain windows, such as primary windows for their development environments. Other users have ran into too many windows that are incompatible with Amethyst, and they would prefer to be choosy with what gets controlled by Amethyst.

## Modifications

* New configuration setting in General > Floating: "Use as Blacklist"
* When "Use as Blacklist" is checked, windows entered into the Floating identifiers will not be automatically set to float (though the "Float small windows" option may still be respected).
* Also, when "Use as Blacklist" is checked, all other windows are floated.

## Video

This video starts with standard behavior. Google Chrome is the only window in the float list. I demonstrate that other windows are not floating, but Google Chrome is floating and unaffected by layout changes.

Then I enable the new option "Use as blacklist", and I demonstrate that not only is Chrome not floating, but all other windows besides Chrome are floating and unaffected by layout changes.

https://cl.ly/0P0p273E1W3X

## Screenshots

image

[Trello Card](https://trello.com/c/BpPMrUwu/292-amethyst-use-floating-bundle-identifiers-as-blacklist)

### DIFF
--- a/Amethyst/Preferences/GeneralPreferencesViewController.swift
+++ b/Amethyst/Preferences/GeneralPreferencesViewController.swift
@@ -16,6 +16,7 @@ final class GeneralPreferencesViewController: NSViewController, CCNPreferencesWi
 
     @IBOutlet var layoutsTableView: NSTableView?
     @IBOutlet var floatingTableView: NSTableView?
+    @IBOutlet weak var floatingLabel: NSTextField!
 
     private var editingFloatingBundleIdentifier = false
 
@@ -37,6 +38,8 @@ final class GeneralPreferencesViewController: NSViewController, CCNPreferencesWi
 
         layoutsTableView?.reloadData()
         floatingTableView?.reloadData()
+
+        updateFloatingLabel()
     }
 
     @IBAction func addLayout(_ sender: NSButton) {
@@ -174,6 +177,14 @@ final class GeneralPreferencesViewController: NSViewController, CCNPreferencesWi
         UserConfiguration.shared.setFloatingBundleIdentifiers(self.floatingBundleIdentifiers)
 
         floatingTableView.reloadData()
+    }
+
+    @IBAction func toggleFloatingLabel(_ sender: NSButton) {
+        updateFloatingLabel()
+    }
+
+    private func updateFloatingLabel() {
+        floatingLabel.stringValue = UserConfiguration.shared.floatingBundleIdentifiersIsBlacklist() ? "Tile Windows:" : "Float Windows:"
     }
 
     func preferenceIdentifier() -> String! {
--- a/Amethyst/Preferences/GeneralPreferencesViewController.swift
+++ b/Amethyst/Preferences/GeneralPreferencesViewController.swift
@@ -16,7 +16,6 @@ final class GeneralPreferencesViewController: NSViewController, CCNPreferencesWi
 
     @IBOutlet var layoutsTableView: NSTableView?
     @IBOutlet var floatingTableView: NSTableView?
-    @IBOutlet weak var floatingLabel: NSTextField!
 
     private var editingFloatingBundleIdentifier = false
 
@@ -38,8 +37,6 @@ final class GeneralPreferencesViewController: NSViewController, CCNPreferencesWi
 
         layoutsTableView?.reloadData()
         floatingTableView?.reloadData()
-
-        updateFloatingLabel()
     }
 
     @IBAction func addLayout(_ sender: NSButton) {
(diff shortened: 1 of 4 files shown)


philippelatulippe:
I'm thrilled to see a PR that allows using Amethyst on a whitelist basis. This will allow me to start using Amethyst again, and I know many people will appreciate this feature.

I'm wondering if this should say "Use as whitelist" instead of "Use as blacklist"? I know you deliberately changed the terminology around, probably to make it fit with existing UI better, but for me it feels more confusing. "Amethyst should only tile these windows" seems more intuitive than "Amethyst should float all windows except these". Perhaps the settings screen should use a separate list UI component for this mode.

axs221:
@philippelatulippe Thanks for the feedback, I can play around with it to make it more intuitive. I think it would be best to reword the items. Having a separate list might not make sense since the settings are mutually exclusive, and it takes up more real estate. I can possibly change the helper text dynamically based on if the check box is checked or not.

philippelatulippe:
I've been been thinking about this, here's a few ideas.

Alternative 1: Call the option "Float all windows except those listed", with a label (with black text) that says "Floated windows:" or "Tiled windows:" depending on whether the option is checked.

Alternative 2: Add a section below the "Floating:" section called something like "Whitelist:". To its right have the setting for "Float all windows". When this option is enabled, a window ID list appears underneath, and the list in the "Floating:" section is hidden.

axs221:
@philippelatulippe I took the option Alternative 1:

image

image

Amethyst/Preferences/UserConfiguration.swift
```diff
@@ -289,20 +291,22 @@ final class UserConfiguration: NSObject {
...
         }
 
+        let useIdentifiersAsBlacklist = floatingBundleIdentifiersIsBlacklist()
+
         for floatingBundleIdentifier in floatingBundleIdentifiers {
             if floatingBundleIdentifier.contains("*") {
                 let sanitizedIdentifier = floatingBundleIdentifier.replacingOccurrences(of: "*", with: "")
                 if runningApplication.bundleIdentifier?.hasPrefix(sanitizedIdentifier) == true {
-                    return true
+                    return useIdentifiersAsBlacklist ? false : true
```
ianyh: Can you just return `!userIdentifiersAsBlacklist` instead of the ternary?

Amethyst/default.amethyst
```diff
@@ -190,6 +190,7 @@
         "key": "x"
     },
     "floating": [],
+    "floating-is-blacklist": false,
```
ianyh: This should probably be true by default as that is the existing behavior.

Amethyst/Preferences/GeneralPreferencesViewController.swift
```diff
@@ -176,6 +179,14 @@ final class GeneralPreferencesViewController: NSViewController, CCNPreferencesWi
         floatingTableView.reloadData()
     }
 
+    @IBAction func toggleFloatingLabel(_ sender: NSButton) {
+        updateFloatingLabel()
+    }
+
+    private func updateFloatingLabel() {
+        floatingLabel.stringValue = UserConfiguration.shared.floatingBundleIdentifiersIsBlacklist() ? "Tile Windows:" : "Float Windows:"
```
ianyh: I think you can do this in Interface Builder by binding a transformation of the key in the user defaults controller. That might be more consistent if you can get it to work.

axs221: Yeah I have something working with a custom transformer. I don't think there are any other custom transformers in the project. Let me know if you want it moved to a different file -- it is pretty specific and probably not going to be used anywhere else, so I kept it in the controller file for now.

Amethyst/Preferences/UserConfiguration.swift
axs221: Yeah, updating

Amethyst/default.amethyst
axs221: Updated

Amethyst/Preferences/GeneralPreferencesViewController.swift
axs221: I followed what was mentioned here to implement the transformer: https://stackoverflow.com/questions/25534334/custom-nsvaluetransformer-in-xcode-6-with-swift

ianyh:
Okay. This generally looks good, but now has some xib conflicts that I'm a bit too tired to deal with right now. 🙂

You can take a look at them yourself if you want, otherwise I'll have a go at resolving them soon.

ianyh:
Conflicts were pretty hairy, but I think I worked it out. I did it on a different branch, though.

Issue #662, opened 2017-11-17
SIWindow.shouldBeManaged() incorrectly returns false during `kAXApplicationShownNotification` handling


ianfixes:
## System

 - **Amethyst version**: 0.11.3
 - **macOS version**: 10.12.6

## What's the problem?
`SIWindow.shouldBeManaged()` has a guard statement for Silica's `SIAccessibilityEvent.isMovable()`.  `isMovable()` returns false in a window's `kAXApplicationShownNotification` observer function.  This happens because `AXUIElementIsAttributeSettable` for `kAXSizeAttribute` returns `kAXErrorAttributeUnsupported` in this instant.  

This is obviously an incorrect result for `shouldBeManaged()`, since windows that were being managed before being hidden do not pass this same check when being shown.

However, the Silica `isMovable()` function is technically correct and should not be changed.  This suggests a more sophisticated `shouldBeManaged()` implementation is needed.

### History list
When windows are hidden, mark them in an set such as "windowsThatWereManaged".  If windows are closed, remove them from this set.  If windows are shown, have `SIWindow.shouldBeManaged()` check the set first and defer to that judgment (then remove them from the set).  The problem here is that there may be legitimate reasons for applications to arbitrarily change their ability to be moved, and this would confuse our check.  

### Arbitrary delay
When windows are shown, set a small delay timer (~15ms) and delay the attempt to `windowManager.addWindow()`. This might require some ability to cancel the addition if (e.g.) the app happens to crash during that time and the window is no longer valid.

[Trello Card](https://trello.com/c/zHKslGe0/299-amethyst-siwindowshouldbemanaged-incorrectly-returns-false-during-kaxapplicationshownnotification-handling)

PR #664, opened 2017-11-21
Respond to application hide/show events, fixes #575

### LINKED ISSUE #575: Relayout on hide
I have two windows open and unmaximized VSCode and Firefox. Amethst will make them side by side. When I do Command + M to minimize one window, it is not focusing the other window. If I do Command + H it does hide the window and focus goes to firefox, however relayout does not take place.

1. Can you please help me to make it so on hide, it does re-layout.
2. Or if you can help me to make Command + M to switch app focus after minimization.

Ideally I would love if Command + M swithced app focus, and if a Command + Tab would unminimize, but it seems Command + Tab doesnt do that to a minimized window. 

[Trello Card](https://trello.com/c/NngvMXLk/271-amethyst-on-minimize-focus-next-window-app-or-relayout-on-hide)

ifreecarve:
This patch adds what appears to be an oversight -- an Amethyst response to application hide/show events.  While testing this patch, I observed that the events seem to fire at the appropriate times, but the handling of `kAXApplicationShownNotification` is sometimes mistakenly aborted.  I've opened #662 to track that, as the proper fix is not immediately obvious to me.

Fixes #575 

[Trello Card](https://trello.com/c/woBCxWMr/300-amethyst-respond-to-application-hide-show-events-fixes-575)

### DIFF
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -187,6 +187,20 @@ private class ObserveApplicationNotifications {
                 windowManager.addWindow(window)
             }
 
+            application.observeNotification(kAXApplicationHiddenNotification as CFString!, with: application) { accessibilityElement in
+                guard let window = accessibilityElement as? SIWindow else {
+                    return
+                }
+                windowManager.removeWindow(window)
+            }
+
+            application.observeNotification(kAXApplicationShownNotification as CFString!, with: application) { accessibilityElement in
+                guard let window = accessibilityElement as? SIWindow else {
+                    return
+                }
+                windowManager.addWindow(window)
+            }
+
             application.observeNotification(kAXFocusedWindowChangedNotification as CFString!, with: application) { _ in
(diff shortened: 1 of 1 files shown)


ifreecarve:
Obligatory one-month bump

Amethyst/Managers/WindowManager.swift
```diff
@@ -187,6 +187,20 @@ private class ObserveApplicationNotifications {
                 windowManager.addWindow(window)
             }
 
+            application.observeNotification(kAXApplicationHiddenNotification as CFString!, with: application) { accessibilityElement in
```
ianyh: Seems like this should be at the application level rather than the window level. If each window is observing then an application visibility change would trigger one event for each window.

ifreecarve: Seeking comments @ianyh ^

ifreecarve: Bump

ianyh: I think i was actually mistaken in my assessment. I think your code is correct.

PR #665, opened 2017-11-21
Don't subscribe to mouse events we don't care about

### LINKED ISSUE #658: Amethyst uses CPU at all times when the mouse is moving
## System

 - **Amethyst version**: 0.11.4
 - **macOS version**: 10.13.1

## What's the problem?
Amethyst uses CPU at all times when the mouse is moving. It's not a ton, but enough to be bothering since it constantly slightly slows down the system.  
Focus follows mouse is **not** enabled.

## How can it be reproduced?
Launch amethyst, open activity monitor and move the cursor.

Amethyst

Not moving cursor.

## dotfile (if you have one)
\-

## Screenshots
\-

ifreecarve:
### Summary
Add an observer to the user settings, and only subscribe to mouse events if we actually want them.

### Test method:
Assumption: focus-follows-mouse is disabled

1. open Activity Monitor
2. open Amethyst
3. wiggle mouse rapidly, note CPU usage of ~0%
4. open preferences
5. wiggle mouse rapidly, note CPU usage of ~3-9% (not sure why this happens)
6. enable focus-follows-mouse
7. close preferences
8. wiggle mouse rapidly, note CPU usage of ~3-9%
9. open preferences, disable focus-follows-mouse
10. close preferences
11. wiggle mouse rapidly, note CPU usage of ~0%

Fixes #658


ifreecarve:
Obligatory one-month bump

mcsosa121:
Obligatory two month bump

ianyh:
This one is very high priority for me. The CPU usage is pretty unacceptable.

Amethyst/Managers/FocusFollowsMouseManager.swift
```diff
@@ -14,21 +14,65 @@ protocol FocusFollowsMouseManagerDelegate: class {
...
-            self.focusWindowWithMouseMovedEvent(event)
+        self.mouseMovedEventHandler = nil
+
+        // must fully init self before we can use it in a closure, so we put in a fake closure for the handler (for now)
+        self.focusConfigurationChangeHandler = { _ in () }
+        super.init()  // fully init self
+
+        // can now set up self as an observer.
+        // we want to observe changes to the focusFollowsMouse config, because mouse tracking has CPU cost
+        UserDefaults.standard.addObserver(self, forKeyPath: ConfigurationKey.focusFollowsMouse.rawValue, options: NSKeyValueObservingOptions.new, context: nil)
```
ianyh: This should probably be behind user configuration. Also, take a look at RxCocoa bindings to `UserDefaults`. It's a bit of a different look at it, but you can do something like
```swift
UserDefaults.standard.rx.observe(...)
    .distinctUntilChanged()
    .scan(nil) { monitor, followingIsDesired in
        if let monitor = monitor {
            NSEvent.removeMonitor(monitor)
            return nil
        } else if followingIsDesired {
            return NSEvent.addGlobalMonitorForEvents(...)
        } else {
            return nil
        }
    }
    .subscribe(...)
```
It takes a bit of grokking the first few times looking at it like this, but it avoids some of the state tracking necessary and is a bit easier to read.

ianyh:
In general a good change, but it's worth working through the more frp way of doing things.

Amethyst/Managers/FocusFollowsMouseManager.swift
```diff
@@ -9,65 +9,45 @@
...
-                }
-                self.mouseMovedEventHandler = NSEvent.addGlobalMonitorForEvents(matching: .mouseMoved) { event in
-                    self.focusWindowWithMouseMovedEvent(event)
-                }
-            } else {
-                guard let handler = self.mouseMovedEventHandler else {
-                    return
+        let stupidCompilerWorkaround: Observable<Any?> = UserDefaults.standard.rx.observe(Bool.self, ConfigurationKey.focusFollowsMouse.rawValue)
+            .distinctUntilChanged { $0 == $1 }
+            .scan(nil) { existingHandler, followingIsDesired in
```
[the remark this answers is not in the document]
ianyh: Additionally, this block captures `self`, but I think you get a retain loop. You store the disposable which means the manager is holding reference to the block which is holding a reference to itself. One way to get around this is to specify that the self reference be either `weak` or `unowned`. The former means that `self` becomes an optional inside the closure which becomes `nil` once `self` is deinitialized. The latter is mostly the same except that it does not become nil so a reference after the fact will segfault. In this case the only reference held is by the manager which disposes it at tear down time so you can safely use unowned, which makes it a bit easier to work with.
```swift
.scan(nil) { [unowned self] existingHandler, followingIsDesired -> Any? in
```

Amethyst/Managers/FocusFollowsMouseManager.swift
```diff
@@ -9,65 +9,45 @@
...
+                if let handler = existingHandler {
+                    NSEvent.removeMonitor(handler)
+                    return nil
+                } else if followingIsDesired! {
+                    return NSEvent.addGlobalMonitorForEvents(matching: .mouseMoved) { event in
```
ianyh: As above you'll need to specify `unowned` here as well as the observable will also be maintaining a reference to this handler.

ifreecarve:
Thanks, fixed.  The compiler errors were extremely unhelpful.

ianyh:
Yeah, the compiler errors for swift are still pretty basic. When in doubt specify types, particularly for closure parameters and return types.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ianfixes:
Monthly bump.  All PR comments were addressed, to my knowledge.

Amethyst/Managers/FocusFollowsMouseManager.swift
```diff
@@ -18,17 +19,39 @@ final class FocusFollowsMouseManager {
...
+                    return nil
+                } else if followingIsDesired! {
+                    return NSEvent.addGlobalMonitorForEvents(matching: .mouseMoved) { [unowned self] event in
+                        self.focusWindowWithMouseMovedEvent(event)
+                    }
+                } else {
+                    return nil
+                }
+            }
+            .subscribe()
```
ianyh: Instead of storing the `Disposable` you should keep a `DisposeBag` and use `disposed(by:)`

ianfixes: I had tried that initially, but ran into a problem with the closure capturing `self`, hence https://github.com/ianyh/Amethyst/pull/665/files#diff-04ea7b5975998885cc0693ef86642bd3R23

If you see an obvious way around that you'll have to be explicit about it :(

ianfixes: I pushed this change, not sure why it worked out better this time :)

ianyh:
It looks like your window peeking change got mixed into this. If you pull that out this should be good to merge.

ianyh:
Awesome!

PR #672, opened 2018-01-15
Maintain required window dimensions regardless of position

ianyh:
I think there may have been some behavioral conflicts between mouse state tracking and minimum window sizes. I've tried to reconcile them in this change, but there is one issue where moving between spaces causes the focused window to jitter and about 50/50 settles in to the wrong position.

@axs221 @ifreecarve curious to get your thoughts. Anything code-wise I might have missed? Ideas on the right behavior?

[Trello Card](https://trello.com/c/exLO8vGU/305-amethyst-maintain-required-window-dimensions-regardless-of-position)

### DIFF
--- a/Amethyst/Layout/Layout.swift
+++ b/Amethyst/Layout/Layout.swift
@@ -52,7 +52,7 @@ struct FrameAssignment {
     var finalFrame: CGRect {
         var ret = frame
         let padding = floor(UserConfiguration.shared.windowMarginSize() / 2)
-        
+
         if UserConfiguration.shared.windowMargins() {
             ret.origin.x += padding
             ret.origin.y += padding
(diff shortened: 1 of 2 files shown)


Amethyst/Layout/Layout.swift
```diff
@@ -88,7 +86,22 @@ struct FrameAssignment {
...
     fileprivate func perform() {
+        var finalFrame = self.finalFrame
+        var finalOrigin = finalFrame.origin
+
+        // Just resize the window first to see what the dimensions end up being
+        // Sometimes applications have internal window requirements that are not exposed to us directly
+        finalFrame.origin = window.frame().origin
+        window.setFrame(finalFrame)
+
+        // If this is the focused window then we need to shift it to be on screen regardless of size
```
ifreecarve: I had to explicitly remove this from the resize-with-mouse code.  The jitter is caused by polluting the event stream -- the move/resize events that we're causing. 

Eventually, I decided that the user is the best person to resolve conflicting constraints (minimum window sizes, frame sizes, etc).  If we owned the entire event stream, it would be a different story.  That's in part due to the fact that attempting to resolve the conflicts myself resulted in 1 of 3 things: 

1. jitter
2. failing to respond to some move/resize events
3. ugly hacks related to setting arbitrary timers

ianyh: Right, right. I knew there was something about that we had discussed. Thank you for the reminder.

ianyh: So it's possible that the correct behavior for now is take the minimum size modifications in the `finalFrame`, but not do the two-pass positioning. I think I will miss the behavior of the windows peeking out from the edge, but I'll have to think more about how that could be accomplished. It seems like what's really missing is a way to track whether or not a space transition is still in the process of animating.

ifreecarve: I too will miss that behavior -- I honestly like the feature of windows attempting to stay on screen, and I'd be glad to hear that there's a way to accomplish it.  It's entirely possible that there are other options out there that I'm failing to consider.  

Being able to track the animation process would be huge.  In general, a way to synchronize the mouse stream with the accessibility events stream (and possibly our own 3rd stream regarding actions in process) would enable a lot of awesome stuff.

Another way to get around the issue would be to make it possible to resize secondary frame boundaries with the mouse.  That's a challenging technical exercise, but well within the realm of "we have control over this" (vs the event streams, which we do not control).

PR #685, opened 2018-02-14
Add peeking of focused window if it's too large for its frame

ianyh:
@ifreecarve could you take a look at this and let me know if I'm missing something? I've been testing it with mouse behaviors enabled and everything seems to work fine. The jittering that I noticed previously was actually a separate issue.

### DIFF
--- a/Amethyst/Layout/Layout.swift
+++ b/Amethyst/Layout/Layout.swift
@@ -86,6 +86,27 @@ struct FrameAssignment {
     }
 
     fileprivate func perform() {
+        var finalFrame = self.finalFrame
+        var finalOrigin = finalFrame.origin
+
+        // If this is the focused window then we need to shift it to be on screen regardless of size
+        if focused {
+            // Just resize the window first to see what the dimensions end up being
+            // Sometimes applications have internal window requirements that are not exposed to us directly
+            finalFrame.origin = window.frame().origin
+            window.setFrame(finalFrame, withThreshold: CGSize(width: 1, height: 1))
+
+            // With the real height we can update the frame to account for the current size
+            finalFrame.size = CGSize(
+                width: max(window.frame().width, finalFrame.width),
+                height: max(window.frame().height, finalFrame.height)
+            )
+            finalOrigin.x = max(screenFrame.minX, min(finalOrigin.x, screenFrame.maxX - finalFrame.size.width))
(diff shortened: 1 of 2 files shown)


ifreecarve:
I'm running this branch on my machine now, and it seems to work.  There's one strange case though: you can accidentally trigger a window swap if you move a peeked window.  Easy to reproduce:

1. Consider a screen that's 1200px wide, and two windows.  Let's say each window has a 600px minimum width.
2. Arrange them in Tall layout, and drag the window on the left to 1000px width.  The window on the right shrinks to its 600px minimum, and is reflowed with 200px on screen and 400px off screen.  
3. Focus the window on the right.  It now overlaps the window on the left by 400px.
4. Put your mouse cursor in the center of the title bar of the window on the right (which should be over the overlapped region) and move the window on the right downward by 2 px

Expected behavior: Dragged window reflows to the peeked position
Actual behavior: Dragged window swaps with window on the left, because technically you dragged it within the bounds of the other frame

I'm not sure the best way to address this.  One would be to make peeking a toggle-able option, and say "yeah when you have window peeking on, the same rules of swapping still apply -- if you drag into another frame then that's where you go.

A second way would be to recognize that a window is peeked, and interpret the move event as a resize event such that the minimum size required by the window you're dragging (the window on the right, in my example) will become the exact frame size.  This might be a nightmare in terms of managing mouse events + accessibility events + focus state

The third way would be to just understand the overlapping that peeking causes, and disable the swapping behavior when in that region.  I'm not precisely sure how to do that if the window has both minimum width and minimum height and you're in a secondary frame.  What do you think?

ifreecarve:
Looking more closely, it seems that the real problem lives in `WindowManager.alternateWindowForScreenAtPoint`.  

Window swaps are calculated by asking what windows are on the screen at a given point.  I think what should be happening instead is asking what _frame_ is supposed to be at that point.  I'm just not sure where to get that information from.

Should we just open a separate issue for `WindowManager.alternateWindowForScreenAtPoint` and merge this now?  The problem predates the code you're adding here.

ianyh:
The idea of expected frame rather than real frame is interesting. It's also not immediately clear to me what the information flow there is, but I am inclined to merge this and track the mouse collision issue separately.

ifreecarve:
Makes sense to me.  This was one of the first parts of the codebase I really messed with, before I knew where things were and what they had access to.  I'll try to crack open that new issue in a week or so.

But yeah, none of that work really blocks merging.  This PR works smoothly and adds back the usefulness that I couldn't deliver.

PR #697, opened 2018-03-20
Added shortcut for real restart of Amethyst

skast96:
added a new mod shortcut for easier use
mod3 -> shift + command
added new function called toggle-amethyst, this calls the restart method from AppDelegate

### DIFF
--- a/Amethyst/Events/HotKeyManager.swift
+++ b/Amethyst/Events/HotKeyManager.swift
@@ -179,6 +179,11 @@ final class HotKeyManager: NSObject {
             windowManager.markAllScreensForReflowWithChange(.unknown)
         }
 
+        constructCommandWithCommandKey(CommandKey.toggleAmethyst.rawValue) {
+            let appDelegate = NSApplication.shared.delegate as! AppDelegate
+            appDelegate.relaunch("Restart" as AnyObject)
+        }
+
         constructCommandWithCommandKey(CommandKey.reevaluateWindows.rawValue) {
             windowManager.reevaluateWindows()
         }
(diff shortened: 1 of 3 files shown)


skast96:
Did anyone check my pullrequest yet ?

ianyh:
I have a couple code comments that I can’t inline right now, but in broad strokes.

- I would like to avoid adding another mod. The dotfile in general is a holdover from the initial days of the app because it was easier to do than a preferences window. I’m torn on where to take it, but right now they are incompatible with preferences set inside the app so something needs to change. Adding another mod feels like an unnecessary increase in complexity. That said, I don’t have a better idea for which command keys to use.
- Accessing the app delegate as a singleton on the app is poor form. Possibly there should be a call to WindowManager which can handle the logic through a delegate instead.

Outside of the code I’m a little worried about having relaunching be so easy. What’s your thinking with it? Is this to force it to grab windows it won’t recognize?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

skast96:
The real reason behind this is to relaunch Amethyst completely because sometimes it´s crashing and not working the way it should.
I sure can rebuild this pull request the right way, as for me and some mates of mine it is a really good feature.

ianyh:
Thinking on it, xmonad does do the relaunch by shortcut thing. If you address the issues we can get this merged.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

max-sixty:
At the risk of "me too" replies on PRs, this would make any small bugs like #838 significantly less painful before they get fixed

ianyh:
Closing in favor of #956

PR #699, opened 2018-03-29
Fix race condition caused by drag events and window peeking

### LINKED ISSUE #690: Window peeking causes a race condition resulting in an unexpected window swap
## System

 - **Amethyst version**:  `0.12.0b`
 - **macOS version**: `10.12.6 (16G29)`

## What's the problem?

Caused by #685 : with window peeking now (re)enabled, windows now move on their own, which creates window move events.  If window focus is changed via a drag event instead of a click event when a window is peeking, an unexpected swap takes place.

## How can it be reproduced?

Tall layout as follows:

```
+---------+-----------+
|         |           |
|         |   Xcode   |
|         |           |
| Chrome  +-----------+
|         |           |
|         |  Terminal |
|         |           |
+---------+-----------+

```

1. Resize Chrome larger and larger until Xcode reaches its minimum width, then go a bit further
2. Focus Xcode so that it peeks
3. Drag to select text in Terminal

Expected behavior: Terminal window regains focus

Actual behavior: 

* `leftMouseDown` event is registered
* OSX causes Xcode window to lose focus
* `kAXFocusedWindowChangedNotification` event is registered
* Amethyst responds to event by unpeeking Xcode window
* `kAXWindowMovedNotification` event is triggered
* `leftMouseDragged` event is registered again because you are dragging
* Amethyst interprets the drag event in Terminal as (mistakenly) being correlated with the move event in Xcode.
* This means that Xcode has been "dragged" to a point within the Terminal window
* Amethyst swaps the windows accordingly.

## Suggested fix

The only one I can think of right now is a delay threshold (sort of hack-y) when unpeeking a window, so that it does not get considered for swaps.  This should be relatively safe, since if you drag the peeked window yourself it should not actually unpeek.  _Hopefully_ these can be considered mutually-exclusive states and we can get away with this.

[Trello Card](https://trello.com/c/p4qJhsjc/313-amethyst-window-peeking-causes-a-race-condition-resulting-in-an-unexpected-window-swap)

ianfixes:
Fixes #690 

This patch adds some event propagation (via closures) so that the mouse state -- used to correlate the mouse event stream with the window event stream -- can be marginally aware of events that were _not_ triggered by the user.

In other words, if window focus is changed by a drag event (triggering a reflow), the movements caused by the reflow would ordinarily cause the `MouseStateKeeper` to think that the drag operation and window move were correlated.  Most commonly, this would cause unexpected window swaps.  

By signaling the beginning (scheduling) and ending (completion) of reflows, we can _de_-correlate mouse and window movements as necessary.  

Please double check my work in regards to reference cycles, I'm inexpert at that.

### DIFF
--- a/Amethyst/Layout/Layout.swift
+++ b/Amethyst/Layout/Layout.swift
@@ -124,7 +124,19 @@ class ReflowOperation: Operation {
         self.frameAssigner = frameAssigner
         self.onReflowCompletion = nil
         super.init()
-        self.completionBlock = nil
+        makeCompletionBlock(nil)
+    }
+
+    private func makeCompletionBlock(_ aBlock: (() -> Void)?) {
+        super.completionBlock = { [unowned self] in
+            aBlock?()
+            guard !self.isCancelled else { return }
+            self.onReflowCompletion?()
+        }
+    }
+
+    public func enqueue(_ aQueue: OperationQueue) {
+        aQueue.addOperation(self)
     }
 
     // Carve out a separate completion block for reflow stuff.
@@ -135,11 +147,7 @@ class ReflowOperation: Operation {
             return super.completionBlock
         }
         set {
-            self.completionBlock = { [unowned self] in
-                newValue?()
-                guard !self.isCancelled else { return }
-                self.onReflowCompletion?()
-            }
+            makeCompletionBlock(newValue)
         }
     }
 
--- a/Amethyst/Layout/Layout.swift
+++ b/Amethyst/Layout/Layout.swift
@@ -124,10 +124,22 @@ class ReflowOperation: Operation {
         self.frameAssigner = frameAssigner
         self.onReflowCompletion = nil
         super.init()
-        self.completionBlock = { [unowned self] in
-            guard !self.isCancelled else { return }
-            guard let onReflowCompletion = self.onReflowCompletion else { return }
-            onReflowCompletion()
+        self.completionBlock = nil
+    }
+
+    // Carve out a separate completion block for reflow stuff.
+    // It will always fire after any existing completion block
+    // UNLESS the operation completed by being cancelled.
+    override public var completionBlock: (() -> Void)? {
+        get {
+            return super.completionBlock
+        }
+        set {
+            self.completionBlock = { [unowned self] in
+                newValue?()
+                guard !self.isCancelled else { return }
+                self.onReflowCompletion?()
+            }
         }
     }
 
--- a/Amethyst/Layout/Layout.swift
+++ b/Amethyst/Layout/Layout.swift
@@ -90,6 +90,7 @@ struct FrameAssignment {
         var finalOrigin = finalFrame.origin
 
         // If this is the focused window then we need to shift it to be on screen regardless of size
+        // We call this "window peeking" (this line here to aid in text search)
         if focused {
             // Just resize the window first to see what the dimensions end up being
             // Sometimes applications have internal window requirements that are not exposed to us directly
@@ -115,12 +116,23 @@ class ReflowOperation: Operation {
     let screen: NSScreen
     let windows: [SIWindow]
     let frameAssigner: FrameAssigner
+    public var onReflowCompletion: (() -> Void)?
 
     init(screen: NSScreen, windows: [SIWindow], frameAssigner: FrameAssigner) {
         self.screen = screen
         self.windows = windows
         self.frameAssigner = frameAssigner
+        self.onReflowCompletion = nil
         super.init()
+        self.completionBlock = {
+            guard !self.isCancelled else { return }
+            guard let onReflowCompletion = self.onReflowCompletion else { return }
+            onReflowCompletion()
+        }
+    }
+
+    deinit {
+        self.onReflowCompletion = nil
     }
 }
 
--- a/Amethyst/Layout/Layout.swift
+++ b/Amethyst/Layout/Layout.swift
@@ -124,7 +124,7 @@ class ReflowOperation: Operation {
         self.frameAssigner = frameAssigner
         self.onReflowCompletion = nil
         super.init()
-        self.completionBlock = {
+        self.completionBlock = { [unowned self] in
             guard !self.isCancelled else { return }
             guard let onReflowCompletion = self.onReflowCompletion else { return }
             onReflowCompletion()
--- a/Amethyst/Managers/ScreenManager.swift
+++ b/Amethyst/Managers/ScreenManager.swift
@@ -19,6 +19,8 @@ final class ScreenManager: NSObject {
     let screenIdentifier: String
     fileprivate weak var delegate: ScreenManagerDelegate?
     private let userConfiguration: UserConfiguration
+    public var onReflowInitiation: (() -> Void)?
+    public var onReflowCompletion: (() -> Void)?
 
     var currentSpaceIdentifier: String? {
         willSet {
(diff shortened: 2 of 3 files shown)


ianfixes:
Monthly bump.  Have been running this revision out of Xcode for the past month and it seems to have eliminated the jitter completely.

Amethyst/Managers/WindowManager.swift
```diff
@@ -223,15 +237,21 @@ private class ObserveApplicationNotifications {
 
                 switch self.mouse.state {
                 case .dragging:
+                    // be aware of last reflow time, again to prevent race condition
+                    guard let delegate = self.mouse.delegate else { break }
+                    let reflowEndInterval = Date().timeIntervalSince(delegate.lastReflowTime as Date)
```
ianyh: You shouldn't need the cast to `Date`. It's already a `Date`.

Amethyst/Managers/WindowManager.swift
```diff
@@ -106,6 +107,19 @@ class MouseStateKeeper {
...
+    // is no longer valid and should be de-correlated from the mouse
+    func handleReflowEvent() {
+        switch self.state {
+        case .doneDragging:
+            self.state = .pointing // remove associated timestamp
+        case .moving:
+            self.state = .dragging // remove associated window
+        default: ()
+        }
+
```
ianyh: Extraneous new line.

Amethyst/Managers/WindowManager.swift
```diff
@@ -620,6 +643,17 @@ final class WindowManager: NSObject, MouseStateKeeperDelegate {
 
             if screenManager == nil {
                 screenManager = ScreenManager(screen: screen, screenIdentifier: screenIdentifier, delegate: self, userConfiguration: userConfiguration)
+                screenManager!.onReflowInitiation = {
+                    self.mouseStateKeeper.handleReflowEvent()
```
ianyh: Retain loop from `self`

Amethyst/Managers/WindowManager.swift
```diff
@@ -620,6 +643,17 @@ final class WindowManager: NSObject, MouseStateKeeperDelegate {
...
+                }
+                screenManager!.onReflowCompletion = {
+                    // This handler will be executed by the Operation, in a queue.  Although async
+                    // (and although the docs say that it executes in a separate thread), I consider
+                    // this to be thread safe, at least safe enough, because we always want the
+                    // latest time that a reflow took place.
+                    self.mouseStateKeeper.handleReflowEvent()
```
ianyh: Retain loop from `self`

Amethyst/Layout/Layout.swift
```diff
@@ -115,12 +116,23 @@ class ReflowOperation: Operation {
...
     let frameAssigner: FrameAssigner
+    public var onReflowCompletion: (() -> Void)?
 
     init(screen: NSScreen, windows: [SIWindow], frameAssigner: FrameAssigner) {
         self.screen = screen
         self.windows = windows
         self.frameAssigner = frameAssigner
+        self.onReflowCompletion = nil
         super.init()
+        self.completionBlock = {
```
ianyh: The operation setting its own completion block feels a little strange. Something that creates a reflow operation would have to know the internals to know that it's not supposed to use the operation's `completionBlock` property. Is there a reason to not just let the screen manager set the completion block directly?

Amethyst/Managers/WindowManager.swift
```diff
@@ -620,6 +643,17 @@ final class WindowManager: NSObject, MouseStateKeeperDelegate {
 
             if screenManager == nil {
                 screenManager = ScreenManager(screen: screen, screenIdentifier: screenIdentifier, delegate: self, userConfiguration: userConfiguration)
+                screenManager!.onReflowInitiation = {
+                    self.mouseStateKeeper.handleReflowEvent()
+                }
+                screenManager!.onReflowCompletion = {
+                    // This handler will be executed by the Operation, in a queue.  Although async
```
ianyh: I think you should probably ensure that this happens on the main queue. If you're worried about ordering of completions you could compare the dates to make sure we only ever take the most recent available date.

ianfixes: I'm not worried about the ordering, I just wanted to address the fact that I'm not providing any thread safety on an assignment operation that could happen in multiple threads.  Since the value is always increasing, and since the fudge factor is in tenths of a second, I think we're safe here.

ianfixes: Also, I'm not sure what's meant by "ensure that this happens on the main queue"

Amethyst/Layout/Layout.swift
ianfixes: I wanted to enforce `guard !self.isCancelled else { return }` in a central place, let me try it a different way.

Amethyst/Managers/WindowManager.swift
ianfixes: fixed

ianfixes:
Bump

Amethyst/Managers/ScreenManager.swift
```diff
@@ -125,8 +133,14 @@ final class ScreenManager: NSObject {
...
         changingSpace = false
-        reflowOperation = currentLayout?.reflow(windows, on: screen)
-        OperationQueue.main.addOperation(reflowOperation!)
+
+        let reflowOperation = currentLayout?.reflow(windows, on: screen)
+        reflowOperation?.onReflowCompletion = { [unowned self] in
+            // propagate the completion handler if we have one
+            self.onReflowCompletion?()
+        }
+        self.onReflowInitiation?()
```
ianyh: No need for `self`.

Amethyst/Managers/ScreenManager.swift
```diff
@@ -125,8 +133,14 @@ final class ScreenManager: NSObject {
...
+        reflowOperation?.enqueue(OperationQueue.main)
```
ianyh: Out of curiosity, is the function here to avoid having to unwrap the `reflowOperation`? If so it would probably be better to just do the conditional unwrap with
```swift
if let reflowOperation = reflowOperation {
    ...
}
```

Amethyst/Managers/WindowManager.swift
```diff
@@ -634,6 +656,17 @@ final class WindowManager: NSObject, MouseStateKeeperDelegate {
 
             if screenManager == nil {
                 screenManager = ScreenManager(screen: screen, screenIdentifier: screenIdentifier, delegate: self, userConfiguration: userConfiguration)
+                screenManager!.onReflowInitiation = { [unowned self] in
```
ianyh: Using `unowned` here is actually unsafe. I think right now the lifetime of the screen manager is tied to the lifetime of the window manager, but that's not a guarantee that is explicitly made. In this case I think it would be better to use `weak` as `self` is not the owner of the object in question; the screen manager owns the object (the block).

Amethyst/Managers/WindowManager.swift
```diff
@@ -634,6 +656,17 @@ final class WindowManager: NSObject, MouseStateKeeperDelegate {
...
+                    self.mouseStateKeeper.handleReflowEvent()
+                }
+                screenManager!.onReflowCompletion = { [unowned self] in
```
ianyh: As above, use `weak` here.

ianfixes:
OK, I think this is all set

ianyh:
Great! Thanks.

PR #707, opened 2018-05-12
Use Floating Bundle Identifiers as Blacklist

ianyh:
Conflict resolved #655

### DIFF
--- a/Amethyst.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
+++ b/Amethyst.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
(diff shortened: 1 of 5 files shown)


ianyh:
Connecting to @axs221's original description:

## Related issues

* Main issue: #171
* Duplicate issue: #250 
* An issue where whitelisting was wanted as a workaround: #200 

## Motivation

Some users have mentioned that they either prefer to use Amethyst only for certain windows, such as primary windows for their development environments. Other users have ran into too many windows that are incompatible with Amethyst, and they would prefer to be choosy with what gets controlled by Amethyst.

## Modifications

* New configuration setting in General > Floating: "Use as Blacklist"
* When "Use as Blacklist" is checked, windows entered into the Floating identifiers will not be automatically set to float (though the "Float small windows" option may still be respected).
* Also, when "Use as Blacklist" is checked, all other windows are floated.

## Video

This video starts with standard behavior. Google Chrome is the only window in the float list. I demonstrate that other windows are not floating, but Google Chrome is floating and unaffected by layout changes.

Then I enable the new option "Use as blacklist", and I demonstrate that not only is Chrome not floating, but all other windows besides Chrome are floating and unaffected by layout changes.

https://cl.ly/0P0p273E1W3X

## Screenshots

image





[Trello Card](https://trello.com/c/BpPMrUwu/292-amethyst-use-floating-bundle-identifiers-as-blacklist)

Issue #683, opened 2018-02-09
Amethyst quits when removing layout that isnt there


natestemen:
## System

 - **Amethyst version**: 0.11.4
 - **macOS version**: 10.13.3

## What's the problem?
Amethyst crashes when clicking "-" in the **Layouts** menu under preferences.

## How can it be reproduced?
Open Preferences and click the button circle below.
![screen shot 2018-02-09 at 2 36 14 pm](https://user-images.githubusercontent.com/12703123/36046396-bfca5c26-0da6-11e8-8368-c5da0a488831.png)

## What applications are involved?
 - Amethyst

## Has anything fixed it, even temporarily?
No

[Trello Card](https://trello.com/c/vJzuN9Nr/311-amethyst-amethyst-quits-when-removing-layout-that-isnt-there)

ianyh:
Is this only when there’s no layout in the list highlighted?

natestemen:
exactly, yes.

ianyh:
Fixed in #708

PR #708, opened 2018-05-15
Fix: Prevent crash when deleting with no row selected

jameszaghini:
For "layouts" & "floating" tables in preferences

https://github.com/ianyh/Amethyst/issues/683

### DIFF
--- a/Amethyst/Preferences/GeneralPreferencesViewController.swift
+++ b/Amethyst/Preferences/GeneralPreferencesViewController.swift
@@ -79,15 +79,11 @@ final class GeneralPreferencesViewController: NSViewController, CCNPreferencesWi
     }
 
     @IBAction func removeLayout(_ sender: AnyObject) {
-        guard let selectedRow = layoutsTableView?.selectedRow, selectedRow < self.layouts.count else {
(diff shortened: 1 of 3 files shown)


Amethyst/Preferences/GeneralPreferencesViewController.swift
```diff
@@ -79,15 +79,11 @@ final class GeneralPreferencesViewController: NSViewController, CCNPreferencesWi
     }
 
     @IBAction func removeLayout(_ sender: AnyObject) {
-        guard let selectedRow = layoutsTableView?.selectedRow, selectedRow < self.layouts.count else {
-            return
-        }
+        guard let selectedRow = layoutsTableView?.selectedRow, selectedRow < self.layouts.count, selectedRow != NSTableView.noRowSelectedIndex else { return }
```
ianyh: I think you want `NSNotFound` rather than using a custom -1 value.

ianyh: Right you are!

ianyh:
Thanks!

PR #714, opened 2018-05-27
Fix an error in determining default float from preferences

ianyh:
FYI @ianfixes


ianfixes:
Good to hear that my sanity was, in fact, intact this whole time :)

PR #747, opened 2018-08-27
Focus the previously focused window when switching screens.

### LINKED ISSUE #703: Switching to a screen with multiple windows changes the focused window in column, tall, or wide layouts
Hi! Thanks for writing Amethyst. I've been using a Mac for work after several years of various tiling WMs (ion3, awesome), and this has me pretty excited.

## System

 - **Amethyst version**: 0.12.0 (57)
 - **macOS version**: 10.13.3

## What's the problem?

Switching screens *to* a screen with `column`, `tall` or `wide`  layout changes the focus of the windows.

I've seen it happen with `fullscreen`, but not consistently.

## How can it be reproduced?

1. On screen 1, create one window of application A.
2. On screen 2, create 2 windows of application B.
3. Set the layout of screen 2 to `tall`.
4. Focus the first-created window of application B.
5. Move focus to screen 1 (via amethyst keyboard shortcut), then move focus to screen 2 (also via amethyst keyboard shortcut). 

I'd expect the focused window on screen 2 to be the previously focused window from step (4). However, regardless of what window was in focus in step (4), amethyst will always focus the same window in step (5).

If the focus doesn't change, focus the *other* window in screen 2, then switch focus back to screen 1, then again to screen 2. The focus should change.

In comparison, if you cmd-tab between the two applications, the focus is preserved.

### Notes

Amethyst pretty reliably picks the first-created window to focus, when screen 2 is in the `tall` layout. It's a different window in `wide` and `column`, but it will be consistent in a given layout.

I have not reproduced this consistently with `fullscreen`.

It also works if you swap screen 1 and 2 above. (E.g. return focus to screen 1.) I don't have more than 2 screens so I can't test with that. :)

I've tested this with:
-  application A = Chrome and application B = Terminal
- application A = Chrome and application B = Chome (same app)

However, my note about the behavior being different if you cmd-tab doesn't apply if application A == application B. 

No other applications are open.

Not that I can find. I tried quitting all other applications, and restarting.

## dotfile (if you have one)

None

## Screenshots

n/a

Thank you!

Lukas-Stuehrk:
Adds tracking of the last focused window per screen. When switching windows, it then tries to focus the previously tracked screen.

Resolves #703

### DIFF
--- a/Amethyst/Managers/ScreenManager.swift
+++ b/Amethyst/Managers/ScreenManager.swift
@@ -17,6 +17,9 @@ protocol ScreenManagerDelegate: class {
 final class ScreenManager: NSObject {
     var screen: NSScreen
     let screenIdentifier: String
+    /// The last window that has been focused on the screen. This value is updated by the notification observations in
+    /// `ObserveApplicationNotifications`.
+    public internal(set) var lastFocusedWindow: SIWindow?
     fileprivate weak var delegate: ScreenManagerDelegate?
     private let userConfiguration: UserConfiguration
     public var onReflowInitiation: (() -> Void)?
--- a/Amethyst/Managers/WindowModifier.swift
+++ b/Amethyst/Managers/WindowModifier.swift
@@ -295,7 +295,7 @@ extension ScreenFocuser {
         }
 
         // If the previous focus has been tracked, then focus the window that had the focus before.
-        if let previouslyFocused = screenManager.lastFocusedWindow {
+        if let previouslyFocused = screenManager.lastFocusedWindow, previouslyFocused.isOnScreen() {
             previouslyFocused.am_focusWindow()
             return
         }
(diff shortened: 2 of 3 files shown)


Amethyst/Managers/WindowModifier.swift
```diff
@@ -294,6 +294,12 @@ extension ScreenFocuser {
             return
         }
 
+        // If the previous focus has been tracked, then focus the window that hat the focus before.
+        if let previouslyFocused = screenManager.lastFocusedWindow {
+            previouslyFocused.focus()
```
ianyh: This needs to be `am_focusWindow()` to get mouse tracking for mouse-follows-focus.

Amethyst/Managers/ScreenManager.swift
```diff
@@ -17,6 +17,9 @@ protocol ScreenManagerDelegate: class {
 final class ScreenManager: NSObject {
     var screen: NSScreen
     let screenIdentifier: String
+    /// The last screen that has been focused on the screen. This value is updated by the notification observations in
```
ianyh: Third word should be window.

ianyh:
Cool. Thanks!

PR #749, opened 2018-08-30
Fix focusing a screen when the screen has no windows.

### LINKED ISSUE #533: Changing Focus Between Screens Fails On Empty Desktop
Currently I have a two-monitor setup and can flip the focus between them with &#8997;&#8679;W and &#8997;&#8679;E (I am also using the `mouse follows focused windows` preference).  This works correctly when I have a window up on each monitor that is being managed by Amethyst.  However, if one of my monitors is currently displaying the desktop and no windows are up, once I move the focus to that monitor it gets stuck and these refocusing shortcuts stop working.  My current solution to this is to either use the application switcher or to move to a different space on the same desktop that might currently have a window open, and once I've moved to that space then focus switching between screens works again.  These workarounds are pretty inconvenient though and definitely slow me down.

Ehlster:
I'm unsure about the naming of the `cgRect` method. Please feel free to suggest a better name.

Resolves #533

### DIFF
--- a/Amethyst/Categories/NSScreen+Amethyst.swift
+++ b/Amethyst/Categories/NSScreen+Amethyst.swift
@@ -36,9 +36,21 @@ extension NSScreen {
         return String(managedDisplay.takeRetainedValue())
     }
 
+    /// Returns the screen's frame translated to Quartz' coordinate system: The screen's frame (and bounds and so on)
+    /// are in cocoa coordinates which are used for views. This coordinate space differs from from Quartz coordinate
+    /// space which is used for events.
+    /// See https://stackoverflow.com/a/19887161 for details.
+    func cgRect() -> CGRect {
+        guard let primaryScreen = NSScreen.screens.first else {
+            return CGRect(origin: frame.origin, size: frame.size)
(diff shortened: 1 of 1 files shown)


ianyh:
The naming seems fine for now, but I tested this and it doesn't seem to solve the problem. That function never gets called. It looks like something in this function is failing to focus the screen when there are no windows. https://github.com/ianyh/Amethyst/blob/development/Amethyst/Managers/WindowModifier.swift#L287

Lukas-Stuehrk:
Interesting....gets called for me at https://github.com/ianyh/Amethyst/blob/development/Amethyst/Managers/WindowModifier.swift#L307 in exactly this method. 

macOS 10.13.6 (17G65) with neither mff nor ffm activated.

ianyh:
So I think it's actually this line https://github.com/ianyh/Amethyst/blob/development/Amethyst/Managers/WindowModifier.swift#L293

There is a focused window, but it's on a different screen. I seem to hit that condition when trying to focus a screen with no windows. I'm not sure why it's only happening to me on your branch.

PR #750, opened 2018-08-30
Add descriptions of layouts to the README

cdisselkoen:
I didn't see the layouts documented anywhere, so I went ahead and documented them all, to the best of my knowledge.  Edits or corrections are welcome.  Also, I'm not sure how the BSP layout works.


README.md
```diff
@@ -67,6 +67,69 @@ And defines the following commands, mostly a mapping to xmonad key combinations.
...
+
+Amethyst allows you to cycle among several different window layouts.
+Layouts can also be enabled/disabled to control whether they appear in the cycle sequence at all.
+
+* *Tall*
+> The default layout. This gives you one "main pane" on the left, and one other
+> pane on the right. By default, one window is placed in the main pane
+> (extending the full height of the screen), and all remaining windows are
+> placed in the other pane. If either pane has more than one window, that pane
+> will be evenly split, horizontally, to show them all.
```
ianyh: I think vertically is more correct than horizontally as it is partitioned along the vertical axis.

README.md
```diff
@@ -67,6 +67,69 @@ And defines the following commands, mostly a mapping to xmonad key combinations.
...
+* *Row*
+> The rotated version of *Column*, where each window takes up an entire row,
+> extending the full width of the screen.
+* *Floating*
+> This mode makes all windows "floating", allowing you to move and resize them
+> as if Amethyst were temporarily deactivated.
+> Unlike the other modes, this will mean that windows can be placed "on top of"
+> each other, obscuring your view of some windows.
+* *Binary Space Partitioning (BSP)*
+> [I'm actually not sure how this one works; feel free to open a PR to explain
```
ianyh: Description of BSP:

This layout does not have a main pane in the way that other layouts do. Each pane can be split into two panes along whatever axis is longer. This is recursive such that pane A can be split in the middle into pane A on the left and pane B on the right; pane B can be split in the middle into pane B on top and pane C on bottom; pane C can be split in the middle into pane C on the left and pane D on the right; and so on.

README.md
cdisselkoen: Hmm, yes, this is confusing.  To me "horizontally" is more intuitive as the layout within the pane resembles the "Row" layout.  Seems readers may disagree on which terminology is more intuitive, so I'll try to be even more explicit to avoid confusion.  This comment also affects the descriptions of some of the other layouts as well.

cdisselkoen: Maybe rather than "split vertically" or "split along its vertical axis", it's better to just stick with "split into rows".

cdisselkoen:
Updated to hopefully clarify the splitting direction, as well as add the description of BSP.

PR #761, opened 2018-10-31
Add 'screen padding' feature

### LINKED ISSUE #469: Allow different margins on edges of screen
It would be great to allow different margins on each edge of the screen vs. what is between windows. This would allow setting a custom margin for the top, left, right, or bottom outside edge of the screen.

The use case for this is as a nice little reserved "tray" like area for desktop icons to remain visible.

[Trello Card](https://trello.com/c/vlRbPPGo/220-amethyst-allow-different-margins-on-edges-of-screen)

deecewan:
I've added the ability to set screen padding. The allows for essentially borders at each edge of the screen.  My use case is that I want to be able to show info bars above and below windows, so I need about 30px offset on the top and bottom, but I don't need any on the left or right, and I don't want to lose 30px off each edge of each window.

Preferences:

image

Before:
image

image

After:
image

image

Note that the subtle changes in between windows is due to the terminal having stepped widths.

Closes #469

### DIFF
--- a/Amethyst/Layout/Layout.swift
+++ b/Amethyst/Layout/Layout.swift
@@ -208,6 +208,11 @@ extension NSScreen {
             frame.size.height = windowMinimumHeight
         }
 
+        frame.origin.x += UserConfiguration.shared.screenPaddingLeft()
+        frame.origin.y += UserConfiguration.shared.screenPaddingTop()
+        frame.size.width -= UserConfiguration.shared.screenPaddingRight()
+        frame.size.height -= UserConfiguration.shared.screenPaddingBottom()
+
         return frame
     }
 }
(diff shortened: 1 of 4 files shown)


ianfixes:
Please test resize-using-mouse with this, to make sure the windows still resize accurately. In other words, that they remain under the mouse cursor when you end the drag operation.

deecewan:
Seems to work as expected:

ianfixes:
Thanks, in particular though I think you'd need to test dragging in the same dimension that's got the extra padding.  In this case, switching to the "wide" layout and resizing vertically would do that.  You should also turn up the edge-of-screen padding to 200px or 300px to magnify any errors in calculation there.

ianyh:
I tested this and it seems to work correctly with mouse dragging and resizing.

Issue #754, opened 2018-09-07
Focus Follows Mouse not working with Google Chrome 69+


qayshp:
## System

 - **Amethyst version**: 0.12.2b3 (64)
 - **macOS version**: 10.13.6

## What's the problem?
FFM no longer switches to Google Chrome windows in Chrome 69+.
If a Chrome window is focused, and you hover over another Chrome window, focus is not moved.
If you are focused on another app and hover over a Chrome window, focus is not moved.
If you have Chrome focused and hover over a window from another app, focus leaves Chrome and you must use a click or the keyboard to move it back.
 
## How can it be reproduced?
Install latest stable Google Chrome and beta release of Amethyst.

## What applications are involved?
Google Chrome v69+

## Has anything fixed it, even temporarily?
Nope.

## dotfile (if you have one)
Nope.

## Screenshots
Nope.

zacharydenton:
I discovered a workaround:

1. Go to <chrome://flags/#views-browser-windows>
2. Change the "Use Views browser windows instead of Cocoa" setting to Disabled.
3. Relaunch Chrome

qayshp:
Thanks! Tried that and it works like a charm again!

Any idea if we should report this as an issue with the feature in Google Chrome, or work on it here, or both?

ianyh:
I think this is probably an Amethyst bug, but I haven’t had a chance to dig in yet.

vespakoen:
The suggested workaround doesn't work anymore, anyone aware of a new workaround for the latest stable google chrome?

zacharydenton:
It looks like the code has been completely removed from Chrome (last update - Jan 9 2019): https://bugs.chromium.org/p/chromium/issues/detail?id=832676

@ianyh Perhaps the CRs linked in that bug provide a hint toward fixing this?

ianyh:
On further investigation the issue actually seems to be window titles not matching. `CGWindowListCopyWindowInfo` reports tab names, but accessibility reports tab names followed with ` - Google Chrome`.

PR #778, opened 2018-12-04
Prefer the use of frame (vs window) boundaries when reacting to mouse drags

ianfixes:
## Problem
This fixes a problem commonly found with the new google chrome where mouse-based swap actions don't actually work. This comes from some strange reporting of window boundaries by google chrome.

### Demo

dec-04-2018 13-26-42 amethyst broken mouse

## Solution

I've slightly tweaked the reflow code to refer to what the layout says the frame boundaries are, not what the windows say their boundaries are.

This has the added effect of being able to swap a window with a fixed-size window like the settings window without dragging your cursor over the settings window itself -- you need only drag the window over the frame occupied by settings.

### Demo

dec-04-2018 13-34-38 amethyst fixed mouse

### DIFF
--- a/Amethyst/Layout/Layout.swift
+++ b/Amethyst/Layout/Layout.swift
@@ -156,6 +156,11 @@ class ReflowOperation: Operation {
     }
 }
 
+// a reflow operation that organizes frames
+protocol FrameReflower {
+    func frameAssignments() -> [FrameAssignment]
+}
+
 protocol FrameAssigner: WindowActivityCache {
     func performFrameAssignments(_ frameAssignments: [FrameAssignment])
 }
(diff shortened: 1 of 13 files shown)


ianyh:
Cool! Does this fix #754?

ianfixes:
I'm hesitant to say that it would fix #754.  The key piece of functionality is here: https://github.com/ianyh/Amethyst/pull/778/files#diff-f9fd72bfb51c345a74f8e484081a933cR287

```swift
// use existing groundwork in layouts / reflowOperations in order to get the window from an assigned frame
extension FramedLayout {
    func windowAtPoint(_ point: CGPoint, of windows: [SIWindow], on screen: NSScreen) -> SIWindow? {
        let assignments = reflowFrames(windows, on: screen).frameAssignments()
        guard let assignment = assignments.first(where: { $0.frame.contains(point) }) else { return nil }
        return assignment.window
    }
}
```

If you tried to hook the focus-folllows-mouse into this technique, I can foresee the following issues:

* It won't work at all in layouts that don't have frames (e.g. floating)
* It will cause strange behavior for fixed-size windows -- they will still focus if the mouse crosses the unoccupied part of their frame
* It might not work on un-managed windows

I guess what I'm really saying here is that it's possible but I don't feel comfortable sorting out all those issues as part of this PR since it isn't a feature I make use of -- I'd be likely to break the experience fo those who use it regularly.

Amethyst/Layout/Layout.swift
```diff
@@ -156,6 +156,11 @@ class ReflowOperation: Operation {
     }
 }
 
+// a reflow operation that organizes frames
+protocol FrameReflower {
+    func frameAssignments() -> [FrameAssignment]
```
ianyh: This could maybe just be on `ReflowOperation` and be optional for the sake of operations that don't have assignments (e.g., floating layout).

ianfixes: I think the reason I put it here is to be able to share code between all layouts that are framed; `FramedLayout` being the glue that allows you to mark a layout as taking advantage of this code.  i.e. if you want to expose the assignments for "all reflow operations that come from a framed layout", I don't think `ReflowOperations` themselves know enough about their corresponding layout types to do this.

ianfixes:
@ianyh I think you'll like these changes -- I was able to remove quite a bit of code by delegating things to the `Layout` and `ReflowOperation` classes.  Much boilerplate has been removed.

I rejected the Hound error because it's complaining about the length of the BSP frame calculation function, which I didn't write :)

PR #780, opened 2018-12-07
Simplify some of the dependency handling

jamie-arcc:
Since the project is already using CocoaPods, SwiftLint should probably be installed that way since a pod is available from it. I added it to the Podfile and took that section out of the setup script.  I also think that, instead of just automatically installing CocoaPods via bundler in the script, we should check for CocoaPods and allow the user/developer to install it how they wish, suggesting to use `gem install cocoapods` as is in the CocoaPods documentation.  I then changed the appropriate line to `pod install` since it would no longer be under bundler.  I left the bundler check in there for now because I am not sure if it is used for anything else in the project.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -13,6 +13,7 @@ target 'Amethyst' do
   pod 'Silica', git: 'https://github.com/ianyh/Silica'
   pod 'Sparkle'
   pod 'SwiftyJSON', '~> 3.1'
+  pod 'SwiftLint'
 
   target 'AmethystTests' do
     inherit! :search_paths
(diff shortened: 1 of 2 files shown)


ianyh:
Using the swiftlint pod seems fine, though you need to update the linting build phase to point to the pod version.

However, I'm skeptical of moving away from bundler. For one thing the Gemfile specifies a specific version which is sometimes relevant when there are beta versions necessary for new Xcode version, for example.

jamie-arcc:
> Using the swiftlint pod seems fine, though you need to update the linting build phase to point to the pod version.
I will fix that.
> However, I'm skeptical of moving away from bundler. For one thing the Gemfile specifies a specific version which is sometimes relevant when there are beta versions necessary for new Xcode version, for example.

Okay.  I got a warning about that actually when I just ran it.

ianyh:
The swiftlint change happened elsewhere and this hasn't seen activity in a long time so I'm going to close it.

PR #813, opened 2019-03-17
Convert the project to use generics for windows and applications

ianyh:
This is an important step towards testability. The Silica objects are very coupled to AX, which makes them really hard to test. Using generics here opens up a lot of opportunity for tests to use mocking. My only concern is that I don't want to add too much of a barrier to entry in terms of contributing to the project.

Hopefully most of the generics complexity is isolated to `Window.swift` and `Application.swift`.


Amethyst/Categories/SIWindow+Amethyst.swift
```diff
@@ -108,7 +122,7 @@ extension SIWindow {
 
         for windowDescription in windowsAtPoint {
             if let window = windowInWindows(windows, withCGWindowDescription: windowDescription) {
-                if window != ignoreWindow {
+                if let ignored = ignoreWindow, window != ignored {
```
ianfixes: Took me a second to parse this. maybe write it as

```swift
  if let ignored = ignoreWindow, ignored != window
```

Amethyst/Categories/SIWindow+Amethyst.swift
```diff
@@ -19,12 +19,26 @@ extension CGRect {
     }
 }
 
-extension SIWindow {
-    // convert SIWindow objects to CGWindowIDs.
+struct WindowsInformation<Window: WindowType> {
```
ianfixes: This should definitely receive some sort of class comment

Amethyst/Events/HotKeyManager.swift
```diff
@@ -68,41 +69,47 @@ final class HotKeyManager: NSObject {
...
+    func setUpWithWindowManager(_ windowManager: WindowManager<Application>, configuration: UserConfiguration) {
         constructCommandWithCommandKey(CommandKey.cycleLayoutForward.rawValue) {
-            windowManager.focusedScreenManager()?.cycleLayoutForward()
+            let screenManager: ScreenManager<Application.Window>? = windowManager.focusedScreenManager()
+            screenManager?.cycleLayoutForward()
         }
 
         constructCommandWithCommandKey(CommandKey.cycleLayoutBackward.rawValue) {
-            windowManager.focusedScreenManager()?.cycleLayoutBackward()
+            let screenManager: ScreenManager<Application.Window>? = windowManager.focusedScreenManager()
```
ianfixes: This logic seems like it should be factored into its own function and replaced with a `guard` here

Amethyst/Managers/ScreenManager.swift
```diff
@@ -138,17 +138,20 @@ final class ScreenManager: NSObject {
         changingSpace = false
 
         let reflowOperation = currentLayout?.reflow(windows, on: screen)
-        reflowOperation?.onReflowCompletion = { [unowned self] in
-            // propagate the completion handler if we have one
-            self.onReflowCompletion?()
+        reflowOperation?.completionBlock = { [weak self, weak reflowOperation] in
```
ianfixes: I like the centralization of this.

Amethyst/Managers/WindowManager.swift
```diff
@@ -13,32 +13,31 @@ import RxSwiftExt
...
 
 // MouseStateKeeper will need a few things to do its job effectively
 protocol MouseStateKeeperDelegate: class {
-    func focusedScreenManager() -> ScreenManager?
-    func windows(on screen: NSScreen) -> [SIWindow]
-    func switchWindow(_ window: SIWindow, with otherWindow: SIWindow)
-    var lastReflowTime: Date { get }
+    associatedtype Window: WindowType
+    func recommendMainPaneRatio(_ ratio: CGFloat)
+    func swapDraggedWindowWithDropzone(_ draggedWindow: AnyWindow<Window>)
```
ianfixes: I assume that this is just delegating the "which window is under the one being dragged" calculation to some other function that we have access to in this scope. Right?

Amethyst/Managers/WindowManager.swift
```diff
@@ -237,48 +213,42 @@ private struct ObserveApplicationNotifications {
...
 
-    private func removeObservers(notifications: [Notification], application: SIApplication) {
-        notifications.forEach { application.unobserveNotification($0.string as CFString, with: application) }
+    private func removeObservers(notifications: [Notification], application: AnyApplication<Application>) {
+        notifications.forEach { application.unobserve(notification: $0.string) }
     }
+}
 
-    private func handle(notification: Notification, element: SIAccessibilityElement, application: SIApplication, windowManager: WindowManager) {
+extension ObserveApplicationNotifications {
```
ianfixes: Class level comment here would be appreciated, to say exactly what kind of thing we're handling and why

Amethyst/Managers/WindowManager.swift
```diff
@@ -334,22 +298,20 @@ private struct ObserveApplicationNotifications {
...
-            guard let screen = resizedWindow.screen(),
-                windowManager.activeWindows(on: screen).contains(resizedWindow) else {
-                    return
-            }
-
-            guard let screenManager = windowManager.focusedScreenManager(),
-                let layout = screenManager.currentLayout as? Layout & PanedLayout,
-                let oldFrame = layout.assignedFrame(resizedWindow, of: windowManager.activeWindowsForScreenManager(screenManager), on: screen) else {
-                    return
+            guard
```
ianfixes: This conditional could benefit from a comment and/or a slight cleanup.  (Maybe some optional chaining?)

Amethyst/Managers/WindowManager.swift
```diff
@@ -523,44 +486,36 @@ final class WindowManager: NSObject, MouseStateKeeperDelegate {
         }
     }
 
-    func reevaluateWindows() {
```
ianfixes: Has this been dead code for a while or was it re-implemented in a place I didn't see?

Amethyst/Model/Application.swift
```diff
@@ -0,0 +1,118 @@
...
+    func pid() -> pid_t
+    func windowWithTitleShouldFloat(_ windowTitle: String) -> Bool
+    func dropWindowsCache()
+    func observe(notification: String, handler: @escaping SIAXNotificationHandler) -> Bool
+    func observe(notification: String, window: AnyWindow<Window>, handler: @escaping SIAXNotificationHandler) -> Bool
+    func unobserve(notification: String)
+    func unobserve(notification: String, window: AnyWindow<Window>)
+}
+
+class AnyApplication<Application: ApplicationType>: ApplicationType {
```
ianfixes: This appears to be a wrapper class. Comment here should explain why we do that (my guess: testability)

Amethyst/Managers/WindowManager.swift
ianfixes: Nevermind, found it

ianyh:
It ended up being an issue with what to generalize. The layouts are parameterized by a window type, but I had trouble getting something generic that didn’t need to be parameterized by a reflow operation type. This would complicate screen manager as it has no way to constrain any particular reflow operation. Making it a concrete type rather than going to the model that Window and Application went to simplified a lot. I talked myself into it by considering what impact it would have on testing. The motivation for window generics was to ease testing, but largely speaking for testing layouts the most interesting things are the reflow operations themselves, which only need a generic frame assigner, and that is already easy.

ianfixes:
I think one of the main drivers of the complexity here is that some of the framed layouts don't support resizing (e.g. BSP), and some of the layouts don't support frames (e.g. floating).  I'm not 100% on the parts of the design that support both of those exceptions to the rule of "frame-based layouts", and I'd love to see maybe a markdown table or some other comparison of what features exist to accommodate those variations.

ianyh:
This is starting to get unwieldy, so I’m going to merge as is. A lot of progress.

ianfixes:
This still looks good

PR #842, opened 2019-07-17
Add focus screens clockwise

kiryph:
Analogue to swap-screen-ccw/cw

### DIFF
--- a/Amethyst/Managers/FocusTransitionCoordinator.swift
+++ b/Amethyst/Managers/FocusTransitionCoordinator.swift
@@ -27,6 +27,8 @@ protocol FocusTransitionTarget: class {
     func screen(at index: Int) -> NSScreen?
     func nextWindowIDClockwise(on screen: NSScreen) -> CGWindowID?
     func nextWindowIDCounterClockwise(on screen: NSScreen) -> CGWindowID?
+    func nextScreenIndexClockwise(from screen: NSScreen) -> Int
+    func nextScreenIndexCounterClockwise(from screen: NSScreen) -> Int
 }
 
 extension FocusTransitionTarget {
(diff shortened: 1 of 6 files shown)


Amethyst/Managers/FocusTransitionCoordinator.swift
```diff
@@ -159,6 +161,90 @@ class FocusTransitionCoordinator<Target: FocusTransitionTarget> {
         // Otherwise focus the topmost window
         target.executeTransition(.focusWindow(topWindow))
     }
+
+    func moveFocusScreenCounterClockwise() {
+        let focusedWindow = Window.currentlyFocused()
```
ianyh: There isn't always a focused window so this can be `nil`, which will crash on the next line. You need to unwrap it or chain from it.

Amethyst/Managers/FocusTransitionCoordinator.swift
```diff
@@ -159,6 +161,90 @@ class FocusTransitionCoordinator<Target: FocusTransitionTarget> {
         // Otherwise focus the topmost window
         target.executeTransition(.focusWindow(topWindow))
     }
+
+    func moveFocusScreenCounterClockwise() {
```
ianyh: This seems to duplicate behavior above in `focusScreen`. Maybe just get the index and call that method instead?

Amethyst/default.amethyst
```diff
@@ -45,6 +45,14 @@
         "mod": "mod1",
         "key": "f"
     },
+    "focus-screen-ccw": {
+        "mod": "mod1",
+        "key": "p"
```
ianyh: I'm wary of decisions on default commands. What's the reasoning here?

kiryph: 1. unused
2. close to home row, i.e. less hand movement (e.g. compared to left/right arrow keys)
3. one hand for modifier, second hand for key
4. mnemonics: next/previous. I also remapped `throw-space-right/left` to `mod2+n/p`. One could use as default `left/right` to be consistent.

Anyhow, right now I use Yabai but keep my eye on Amethyst because I am not the biggest fan of bsp (maybe my displays are not big enough). Therefore I do not work on this PR as long as I stay with Yabai. My experience with Yabai or Amethyst is that macOS is not the best OS for tiling window manager.

ianyh: That's a pretty reasonable explanation. I'll take over the PR to clean it up and get it merged. Thanks for the work so far.

> My experience with Yabai or Amethyst is that macOS is not the best OS for tiling window manager.

A painful truth.

kiryph: > I'll take over the PR to clean it up and get it merged. Thanks for the work so far.

That's great. Thank you.

> A painful truth.

Linux becomes more and more appealing. Unfortunately, I have bought a non-cross platform license of Mathematica for macOS. Also other things keep me from moving away (e.g. long term 1Password user without the new subscription based model).

Anyhow, I really appreciate projects such as Amethyst. Keep up the excellent work!

PR #848, opened 2019-08-13
update cocoapods version

kejadlen:
The Podfile was generated with 1.6.2, so this aligns the minimum
required CocoaPods version to match that.


ianyh:
A strange oversight! I must have accidentally used my not bundled version. Thanks!

PR #854, opened 2019-08-20
Respect mouse follow focus on window swaps and throws (Fixes #853)

### LINKED ISSUE #853: Mouse doesn't follow window when moving windows
## System

 - **Amethyst version**: 0.14.0
 - **macOS version**: 10.14.6

## What's the problem?

I think this was lost at some point when this was converted to Swift? There's still a lone focusedWindow.focus() in WindowTransitionCoordinator.swapFocusedWindowClockwise (but not in the other methods), but I think that solution no longer works due to the async nature of the new code. If you add focusedWindow.focus() calls to the other window transition methods, the mouse cursor gets updated to the last position before the transition, since the focus() call is sync and happens before the reflow.

I tried calling Window.currentlyFocused()?.focus() inside the reflow operation's completionBlock but that leads to some weird behavior where the focus is stuck to the currently focused window and you can't change it.

I think there's an additional bug that may be influencing this behavior in Window.isFocused. I set some breakpoints and inspected the values during reflow operations, and basically it always returns false. Changing the isEqual(to: focused) call to focused.windowID() == self.windowID() seems to do the right thing, or at least returns true when I expect it to.

See also #120 and #147.

reuben:
This fixes #853 for me but I'm not super confident about this approach. I only use paned layouts so don't know how this interacts with more complicated setups.

### DIFF
--- a/Amethyst/Managers/ScreenManager.swift
+++ b/Amethyst/Managers/ScreenManager.swift
@@ -124,6 +124,21 @@ final class ScreenManager<Window: WindowType>: NSObject {
             }
 
             self?.onReflowCompletion?()
+            // If this reflow was caused by a window swap, re-focus the window
+            // so that mouse follows focus is respected if it is set
+            switch event {
+            case let .windowSwap(first, second):
+                if first.isFocused() {
+                    first.focus()
+                } else if second.isFocused() {
+                    second.focus()
+                }
+            case let .add(window):
+                if window.isFocused() {
+                    window.focus()
+                }
+            default: ()
+            }
         }
         onReflowInitiation?()
         if let reflowOperation = reflowOperation {
--- a/Amethyst/Managers/WindowTransitionCoordinator.swift
+++ b/Amethyst/Managers/WindowTransitionCoordinator.swift
@@ -109,7 +109,6 @@ class WindowTransitionCoordinator<Target: WindowTransitionTarget> {
         let windowToSwapWith = windows[(focusedWindowIndex + 1) % windows.count]
 
         target.executeTransition(.switchWindows(focusedWindow, windowToSwapWith))
-        focusedWindow.focus()
     }
 
     func throwToScreenAtIndex(_ screenIndex: Int) {
--- a/Amethyst/Model/Window.swift
+++ b/Amethyst/Model/Window.swift
@@ -165,7 +165,7 @@ extension AXWindow: WindowType {
             return false
         }
 
-        return isEqual(to: focused)
+        return focused.windowID() == self.windowID()
     }
 
     /**


reuben:
I think this has the unintended consequence of snapping the mouse pointer to the middle of a window when e.g. a minimized window is un-minimized, due to the second check added to `ScreenManager` on an .add event. It's meant to handle throwing windows to a different screen, but it's having this additional side effect. I guess the better way to fix it would be to maybe separate change types into .add and .throw, maybe? Then we could skip the re-focus on add.

Amethyst/Model/Window.swift
```diff
@@ -165,7 +165,7 @@ extension AXWindow: WindowType {
             return false
         }
 
-        return isEqual(to: focused)
+        return focused.windowID() == self.windowID()
```
ianyh: Oh, this is interesting. I wonder if there are other bugs stemming from `isEqual` in `SIAccessibilityElement` doing a direct ref comparison. Either way, though, it is probably better to override `isEqual` in `AXWindow` to make sure the behavior is uniform across other comparisons.

Amethyst/Managers/ScreenManager.swift
```diff
@@ -124,6 +124,21 @@ final class ScreenManager<Window: WindowType>: NSObject {
             }
 
             self?.onReflowCompletion?()
+            // If this reflow was caused by a window swap, re-focus the window
+            // so that mouse follows focus is respected if it is set
+            switch event {
```
ianyh: I have to think about this more, but I would like to keep window focusing isolated to the relevant transition coordinator.

Amethyst/Managers/WindowTransitionCoordinator.swift
```diff
@@ -109,7 +109,6 @@ class WindowTransitionCoordinator<Target: WindowTransitionTarget> {
         let windowToSwapWith = windows[(focusedWindowIndex + 1) % windows.count]
 
         target.executeTransition(.switchWindows(focusedWindow, windowToSwapWith))
-        focusedWindow.focus()
```
ianyh: Nice catch.

reuben:
Hey @ianyh, sorry for the delay in responding, things have been a bit hectic. I can separate the two simpler changes (override `isEqual`, remove the `focus()` call) into another PR so they can be merged while you decide what to do about the rest. But I'm not sure if you're separately tacking that elsewhere, since this was put in a "Doing" state 5 days ago. Let me know!

ianyh:
@reuben I moved it to Doing because _you_ are working on it. 🙂

Pulling out the smaller changes sounds good.

reuben:
Created #890 and #891

michaellatman:
Seems to also resolve https://github.com/ianyh/Amethyst/issues/989. I am using this full time now until I can reproduce. The intellij IDEs seem bad at this, a small dialog throws windows around.

ianyh:
Closing in favor of #1013.

PR #870, opened 2019-09-09
Remove AX app shown/hidden notifications from observation

### LINKED ISSUE #662: SIWindow.shouldBeManaged() incorrectly returns false during `kAXApplicationShownNotification` handling

### LINKED ISSUE #741: Hiding window with cmd+h breaks tiling for this window when reappearing.
## System
 - **Amethyst version**:
0.21.1 (61)
 - **macOS version**:
10.13.6 (17G65)

## What's the problem?
After pressing `cmd+h` to hide a window (not minimize), the window is hidden and the remaining windows are tiled correctly. But after opening the closed window with `cmd+tab` the windows do not get rearranged. Forcing with `cmd+z` works though.

ianyh:
There were two issues here.
* We were duplicating logic for hide/unhide. The observed AX event and the `NSWorkspace` notification both triggered when hiding and unhiding an app.
* Applications were keeping bad windows cache when hiding and unhiding.

Fixes #741 
Closes #662


ianfixes:
Reliable/unreliable looks neat.  Is that a common pattern or just something invented here?

ianyh:
It's something I came up with this specifically, but it's pretty close to other models. It's just assigning a type to a binary state. `Optional`, for example, follows this model.

ianfixes:
I really like it, but it's light on comments.  I was hoping that you could shed some light (via added comments) on what is considered unreliable/reliable, and why.  Similarly, some discussion of how to make something unreliable into something reliable would be valuable to future me.

PR #891, opened 2019-09-23
Override isEqual to use window ID instead of pointer equality

reuben:


Amethyst/Model/Window.swift
```diff
@@ -179,6 +179,10 @@ extension AXWindow: WindowType {
         return false
     }
 
+    func isEqual(to rhs: SIWindow) -> Bool {
```
ianyh: `rhs` should `AXWindow`.

reuben: I had done that at first, but then the function was never called from `isFocused`. I now realize it's probably because that was previously calling a non-wrapped SIWindow method. I pushed a fix.

PR #897, opened 2019-10-26
Add Tall Stack layout

cgtz:
Add a layout where windows are stacked atop each other on the left (main) side of the screen and split evenly into rows on the left side.

This is meant to emulate a layout that I use on I3, where I have a tabbed container on the left and a split container on the right side of the screen.


ianyh:
Can’t look at it right now, but could you add tests for this?

cgtz:
sure, I will look into writing a test for this.

PR #910, opened 2019-11-18
[Feature] Add the relaunch shortcut key

20015jjw:
Hi, I'm writing this PR as I find myself manually relaunching Amethyst pretty often, so I guess a shortcut is reasonable. I've been using it for the past 2 years and it's been amazing, so I thought I'd contribute. 
I'm really new to swift / macOS development, so let me know if anything is missing. Thanks.

For this PR, I'm essentially holding a reference to the main app delegate, where the original relaunch method lives, and call it in the hot key manager. I think it's fairly straightforward.

I've also manually verified the feature. It works for me. 
I did, however, need to remove the existing Amethyst entry in the system accessibility permission list and re-request the permission to make it work when I first started to make this.

### DIFF
--- a/Amethyst/AppDelegate.swift
+++ b/Amethyst/AppDelegate.swift
@@ -55,7 +55,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         preferencesWindowController?.window?.level = .floating
 
         windowManager = WindowManager(userConfiguration: UserConfiguration.shared)
-        hotKeyManager = HotKeyManager(userConfiguration: UserConfiguration.shared)
+        hotKeyManager = HotKeyManager(appDelegate: self, userConfiguration: UserConfiguration.shared)
 
         hotKeyManager?.setUpWithWindowManager(windowManager!, configuration: UserConfiguration.shared)
     }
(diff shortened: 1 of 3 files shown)


20015jjw:
Updated to use weak reference to the delegate.

20015jjw:
Oh hum didn't see #697 before I started. I guess I just need a easier access to the existing relaunch feature.

Amethyst/AppDelegate.swift
```diff
@@ -55,7 +55,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         preferencesWindowController?.window?.level = .floating
 
         windowManager = WindowManager(userConfiguration: UserConfiguration.shared)
-        hotKeyManager = HotKeyManager(userConfiguration: UserConfiguration.shared)
+        hotKeyManager = HotKeyManager(appDelegate: self, userConfiguration: UserConfiguration.shared)
```
ianyh: I would pass the delegate in to the setup rather than on initialize.

ianyh:
Made some minor changes in a different PR to get this merged. #956

PR #953, opened 2020-02-21
Make install script more robust to uninstalled rbenv or ruby

ianfixes:
I'm not sure if these fixes are helpful, but I made them to overcome issues with a minimally-configured ruby environment.


ianyh:
Seems reasonable.

PR #996, opened 2020-03-24
Fix layout preferences constraints

greedy:
This pull request fixes a UI bug that occurs when enough items are added to the layout list in the layouts preference pane.

Some of the layout constraints were placed on the table inside of the scrolled view instead of the scrolled view itself. This causes the layout constraints to be unsatisfiable breaking the UI.

The first commit moves the constraints from the table to the scrolled view. The second commit removes the extraneous constraint that caused the layout to be unsatisfiable.


ianyh:
Should fix #999.

PR #1013, opened 2020-05-04
Focus-follows-mouse/mouse-follows-focus improvements

### LINKED ISSUE #853: Mouse doesn't follow window when moving windows

### LINKED ISSUE #1006: "Mouse follows focused windows" shouldn't move the mouse while the user is moving the mouse
Focusing a window with the mouse (either by clicking or using "Focus follows mouse") causes the mouse to jump to the newly focused window. This is not needed since the mouse is already over the newly focused window, and conflicts with the user's actions.

rickyz:
Fixes #1006, fixes #853.

### DIFF
--- a/Amethyst/Managers/ScreenManager.swift
+++ b/Amethyst/Managers/ScreenManager.swift
@@ -12,6 +12,8 @@ import Silica
 protocol ScreenManagerDelegate: class {
     associatedtype Window: WindowType
     func activeWindowSet(forScreenManager screenManager: ScreenManager<Self>) -> WindowSet<Window>
+    func onReflowInitiation()
+    func onReflowCompletion()
 }
 
 final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
@@ -31,8 +33,6 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
     /// `ObserveApplicationNotifications`.
     private(set) var lastFocusedWindow: Window?
     private let userConfiguration: UserConfiguration
-    var onReflowInitiation: (() -> Void)?
-    var onReflowCompletion: (() -> Void)?
 
     private var reflowOperation: Operation?
 
@@ -122,11 +122,6 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
         try values.encode(layoutsBySpaceUUID, forKey: .layoutsBySpaceUUID)
     }
 
-    deinit {
-        self.onReflowInitiation = nil
-        self.onReflowCompletion = nil
-    }
-
     func updateScreen(to screen: Screen) {
         self.screen = screen
     }
@@ -202,9 +197,9 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
                 return
             }
 
-            self?.onReflowCompletion?()
+            self?.delegate?.onReflowCompletion()
         }
-        onReflowInitiation?()
+        delegate?.onReflowInitiation()
         OperationQueue.main.addOperation(reflowOperation)
     }
 
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -189,6 +189,7 @@ final class WindowManager<Application: ApplicationType>: NSObject, Codable {
             return
         }
         markScreen(screen, forReflowWithChange: .unknown)
+        doMouseFollowsFocus(focusedWindow: focusedWindow)
     }
 
     @objc func applicationDidLaunch(_ notification: Notification) {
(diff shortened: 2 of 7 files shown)


Amethyst/Managers/ScreenManager.swift
```diff
@@ -202,9 +197,9 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
                 return
             }
 
-            self?.onReflowCompletion?()
+            self?.delegate?.onReflowCompletion()
         }
-        onReflowInitiation?()
+        self.delegate?.onReflowInitiation()
```
ianyh, line 202: Unnecessary `self`.

Amethyst/Model/Window.swift
```diff
@@ -270,23 +270,7 @@ extension AXWindow: WindowType {
...
-        let windowFrame = frame()
-        let mouseCursorPoint = NSPoint(x: windowFrame.midX, y: windowFrame.midY)
-        guard let mouseMoveEvent = CGEvent(mouseEventSource: nil, mouseType: .mouseMoved, mouseCursorPosition: mouseCursorPoint, mouseButton: .left) else {
-            return true
-        }
-        mouseMoveEvent.flags = CGEventFlags(rawValue: 0)
-        mouseMoveEvent.post(tap: CGEventTapLocation.cghidEventTap)
-
-        return true
+        return super.focus()
```
ianyh, line 273: This override is no longer necessary.

Amethyst/Managers/WindowManager.swift
```diff
@@ -85,24 +85,11 @@ final class WindowManager<Application: ApplicationType>: NSObject, Codable {
     }
 
     func screenManager(screen: Screen) -> ScreenManager<WindowManager<Application>> {
-        let screenManager = ScreenManager<WindowManager<Application>>(
+        return ScreenManager<WindowManager<Application>>(
```
ianyh, line 88: This whole function is only used in one place, I think, so we can probably get rid of it.

Amethyst/Managers/WindowManager.swift
```diff
@@ -407,6 +395,47 @@ final class WindowManager<Application: ApplicationType>: NSObject, Codable {
         // If we've reached this point we haven't found any tab to switch out, but this window could still be new.
         add(window: window)
     }
+
+    func onReflowInitiation() {
+        self.mouseStateKeeper.handleReflowEvent()
```
ianyh, line 400: No `self`.

Amethyst/Managers/WindowManager.swift
```diff
@@ -407,6 +395,47 @@ final class WindowManager<Application: ApplicationType>: NSObject, Codable {
...
+    }
+
+    func onReflowCompletion() {
+        if let focusedWindow = Window.currentlyFocused() {
+            self.doMouseFollowsFocus(focusedWindow: focusedWindow)
```
ianyh, line 405: No `self`.

Amethyst/Managers/WindowManager.swift
```diff
@@ -407,6 +395,47 @@ final class WindowManager<Application: ApplicationType>: NSObject, Codable {
...
+        }
+
+        // This handler will be executed by the Operation, in a queue.  Although async
+        // (and although the docs say that it executes in a separate thread), I consider
+        // this to be thread safe, at least safe enough, because we always want the
+        // latest time that a reflow took place.
+        self.mouseStateKeeper.handleReflowEvent()
```
ianyh, line 412: No `self`.

Amethyst/Managers/WindowManager.swift
```diff
@@ -407,6 +395,47 @@ final class WindowManager<Application: ApplicationType>: NSObject, Codable {
...
+        self.lastReflowTime = Date()
```
ianyh, line 413: No `self`.

Amethyst/Managers/WindowManager.swift
rickyz: To do this, I needed to replace userConfiguration with UserConfiguration.shared - hope that's all right - I don't think any tests rely on injecting a special configuration into ScreenManager.

PR #1103, opened 2021-02-24
Adds relaunch button in layout preference view

suarez-agu:
Simply adds a "Relaunch Amethyst" button to the Layout Manager Preference layout as it is easier and quicker to restart while testing different layouts.

Button enables when there is a change in the layouts selection.
I created the AppManager class to store the relaunch code since it is now reused in two locations

### DIFF
--- a/Amethyst/Preferences/LayoutsPreferencesViewController.swift
+++ b/Amethyst/Preferences/LayoutsPreferencesViewController.swift
@@ -12,7 +12,7 @@ class LayoutsPreferencesViewController: NSViewController, NSTableViewDataSource,
     private var layoutKeys: [String] = []
 
     @IBOutlet var layoutsTableView: NSTableView?
-    @IBOutlet weak var relaunchBtn: NSButton?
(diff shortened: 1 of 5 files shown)


Amethyst/Preferences/LayoutsPreferencesViewController.swift
```diff
@@ -12,6 +12,7 @@ class LayoutsPreferencesViewController: NSViewController, NSTableViewDataSource,
     private var layoutKeys: [String] = []
 
     @IBOutlet var layoutsTableView: NSTableView?
+    @IBOutlet weak var relaunchBtn: NSButton?
```
ianyh: It looks like the xib connects to `relaunchButton`, not `relaunchBtn`. I couldn't get the UI to work without renaming it to the former.

ianyh (changes requested):
Good idea.

suarez-agu:
Apologies for the delays. Thank you for making those changes and merging

PR #1124, opened 2021-06-14
Add TwoPane layout

### LINKED ISSUE #894: TwoPane layout
Hi, would it be possible to implement a [TwoPane](https://hackage.haskell.org/package/xmonad-contrib-0.15/docs/XMonad-Layout-TwoPane.html) layout from XMonad? This layout is very similar to the Tall layout, except that there is only one window showed on the secondary pane and any additional windows in this layout can be cycled through (on the secondary pane).

mwz:
This PR adds a new `TwoPane` layout as discussed in #894.
I've adapted the code from the `Tall` layout and removed the ability to resize the remaining windows on the secondary pane so that there is always only one window visible there. See a quick demo below:

https://user-images.githubusercontent.com/1190768/121961818-d095d680-cd5f-11eb-92e3-3eb180bbe548.mp4

I had to bump some dependencies in order to get the project to compile locally + I had to apply [this](https://github.com/Quick/Quick/issues/1025#issuecomment-821863103) fix to Quick to get the tests running, but you might be able to get away without those changes, so let me know if you want me to revert 3096822.

Closes #894


pomdtr:
This is transformative, thank you so much!

To elaborate, it allows me to work without adding any extra spaces (which is awkward using amethyst).
I can add any number of windows to my main space without running out of monitor room.

ianyh (approved):
Cool!

nadeemkhedr:
really looking forward to this, when will we have a release? (btw just started using amethyst, amazing tool!)

yangwao:
Wow, just received an update, amazing. Love to sponsor this project to keep it rolling!:)

paveltyk:
Thanks. Use it a lot on my 15" display.

diocletiann:
I've been waiting for this, thank you! Would it be possible to crate a variation that allows  alternating the placement of new windows in both panes? That way the last focused window is never obscured. When both stacks fill up you could switch back to this layout and the focused window could go to the main pane.

mwz:
There's a shortcut for setting the currently focused window as the main window, so you can do that before you switch to this layout to have the current window become your main, unless I've misunderstood what you meant - not entirely sure what you mean by "both stacks".

diocletiann:
Sorry if I didn't explain clearly. Basically I'm looking at both panes as stacks of windows, similar to how i3 and yabai do it. Your layout will stack all new windows on top what is in the right pane even if it's currently in focus, obscuring the existing window. This is already the best layout in the app IMO. 

I'm proposing a behavior where the next window you open stacks on top of the pane not in focus, right or left. That way what you were just reading before you launched a new window is still visible and you don't have to move anything.

mwz:
ok yeah that makes sense - the current implementation was inspired by how the TwoPane works in XMonad, but what you're proposing is a variation of this with a different behaviour for the main pane essentially treating it the same as the other pane. That seems to be already tracked in #1130.

diocletiann:
Yes, with the difference being that #1130 still requires either manual selection of the focused pane or manual movement of the new window (the latter is currently not even possible with the keyboard afaik). Placing the new window in the unfocused pane (similar to Emacs `other-window`) solves for both in most situations.

allanalvaro:
The best layout. Can we have the TwoPane-Right? Like we have in the Tall?

mwz:
> The best layout. Can we have the TwoPane-Right? Like we have in the Tall?

Should be straightforward to do - see #1182 for the latest changes to the TwoPane layout. Probably best to use that as a reference if you want to contribute the changes.

PR #1151, opened 2021-09-26
Bump version to 0.15.5

ianyh:


yangwao:
How I can sponsor this? 
At first, I'm using Obsidian where the logo is similar and I was grasping if I'm reading it right that Amethyst has pushed an update. I was wishing for that for a while, here we go! :)

PR #1166, opened 2021-11-19
add regex support to floating windows title matching

### LINKED ISSUE #981: Regular expression support for window title filters
Can I only floating the "Get Info" popup of Finder? I have set it in perference and add `*简介*`(means 'Info' in chinese) in window title restriction, but it will not take effect.
截屏2020-03-1219 45 13

SebastienWae:
Fixes: https://github.com/ianyh/Amethyst/issues/981

Add regex support to the current matching of floating windows title. 
This shouldn't break current configurations.

### DIFF
--- a/Amethyst/Preferences/UserConfiguration.swift
+++ b/Amethyst/Preferences/UserConfiguration.swift
@@ -418,7 +418,13 @@ class UserConfiguration: NSObject {
         // If the title matches it is included
         //   - Blacklist means floating
         //   - Whitelist means not floating
-        if floatingBundle.windowTitles.contains(title) {
+        if floatingBundle.windowTitles.contains(where: { windowTitle in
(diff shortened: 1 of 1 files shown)


ianyh:
Cool!

usmajeed:
Does this work? I haven't been successful in getting my Microsoft Edge DevTools window to float like this.

kerry-iclasspro:
> Does this work? I haven't been successful in getting my Microsoft Edge DevTools window to float like this.

Works for me.  For instance, I want Chrome DevTools to always float, so I set **Window Title Restrictions** to `/DevTools.*/`

kerryj89:
> Does this work? I haven't been successful in getting my Microsoft Edge DevTools window to float like this.

It works for me, in the Window Title Restrictions I just had to put `DevTools*`

Issue #856, opened 2019-08-29
Moving windows between spaces on second screen behaves unexpectedly


zmheiko:
## System

 - **Amethyst version**: 0.14.1b
 - **macOS version**: 10.14.5

## What's the problem?
With a two screen setup trying to move a window on the second screen from space **a** to space **b** results in the window moving to space **b + (number of spaces on the first screen)**.  
Basically the spaces on the second screen are indexed starting from 1 again. Moving between spaces still works as usual, i.e. it uses a joint set of indices to refer to spaces on both screens.
(I know this sounds confusing please see my examples below)

This issue popped up for me when I updated to a version of amethyst that uses the new private APIs for moving windows between spaces.

## How can it be reproduced?
**My setup:**
Number of screens: **2**
Spaces on first screen: **{1,2,3}**
Spaces on second screen: **{4,5,6,7,8,9,10}**

**Examples:**
Trying to move a window from any space on the second screen to **space 1** results in window moving to **space 4 = 1+3**.
Trying to move a window from any space on the second screen to **space 2** results in window moving to **space 5 = 2+3**
...
Trying to move a window from any space on the second screen to **space 7** results in window moving to **space 10 = 7+3**.
Trying to move a window from any space on the second screen to **space 8, 9, 10** results in window not moving.

## What applications are involved?
All applications and windows.

## Has anything fixed it, even temporarily?
Sadly, no.

PR #1209, opened 2022-03-01
Throw window to any space across multiple screens.

dozzman:
This PR adjusts the behaviour of throwing windows between spaces such that:
- You can throw a window to another space by using the space/desktop number visible from Mission Control.
- Throwing windows between spaces on different screens now works as expected (least surprise, matching XMonad behaviour).

Currently throwing windows between spaces only works between spaces on the same screen. You cannot throw a window to a space belonging to a different screen. This is confusing when managing windows in a multi-screen setup i.e. instead of throwing a window to space X (where X is the Mission Control space number) which exists on screen A directly, you must first throw the window to screen A and then throw it to space X' which is the index of space X local to screen A. It creates more steps when issuing keyboard commands as explained in the example. Its also confusing from the fact that the space numbers shown in Mission Control do not reflect the space numbers used when throwing windows in a multi-screen setup. Finally the current behaviour deviates from the existing XMonad behaviour which allows throwing windows to arbitrary workspaces, regardless of which screen it is currently visible on.

### DIFF
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -642,7 +642,7 @@ extension WindowManager: WindowTransitionTarget {
         case let .moveWindowToSpaceAtIndex(window, spaceIndex):
             guard
                 let screen = window.screen(),
-                let spaces = CGSpacesInfo<Window>.spacesForScreen(screen, includeOnlyUserSpaces: true),
+                let spaces = CGSpacesInfo<Window>.spacesForAllScreens(includeOnlyUserSpaces: true),
                 spaceIndex < spaces.count
             else {
                 return
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -649,9 +649,16 @@ extension WindowManager: WindowTransitionTarget {
             }
 
             let targetSpace = spaces[spaceIndex]
-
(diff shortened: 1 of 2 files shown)


dozzman:
I think adding this behaviour makes sense but I appreciate it would break the existing behaviour for folks as a result. I think the best thing is to hide the behaviour behind a flag in preferences but I haven't looked into how to do that. I'd appreciate if anyone with experience in modifying the preferences code could make that change.

dozzman:
I also need to reflow the recipient space as it will not tile correctly if the space is already visible on a screen. Will drop an update shortly.

dozzman:
I've sorted reflow when moving windows between spaces on different screens.

ianyh:
Very cool. Much appreciated contribution. I'll think about how it fits into settings.

ianyh (approved):

schulzf:
how do I make use of this?

I suppose this PR solves my problem:

I just installed and I have 10 spaces open, spaces 1 and 2 on screen 1, and spaces 3 to 10 in screen 2.
when i use `Throw focused window to space 1` on a window in space 5, it goes to space 3....

schulzf:
this is neat, I just ran the development branch, it makes a world of difference for me
When will this be released @ianyh ?

zmheiko:
Great! This also fixes issue [#856](https://github.com/ianyh/Amethyst/issues/856) (closed it).

PR #1241, opened 2022-05-14
Fix swapping with two windows visible for two-pane layout

### LINKED ISSUE #1234: "Move focus to main window" should also move focus back
When working in TwoPane mode I'd like to be able to easily switch between the main pane and the stack.

There is a shortcut for "Move focus to main window" which does nothing if the focus is already on the main window. This shortcut should instead switch focus as per above.

Example: I have window A in the main pane and windows B and C in the right pane (the stack) with B on top. A and B are now visible on the screen. "Move focus to main window" moves focus between A and B. C is not affected and remains hidden behind B.

Tried using both the "swap focus" and the "move focus" shortcuts, none of those do what I want. 

**Additional context**
N/A

If this would be an acceptable solution I could submit a PR.

rickythefox:
Closes #1234 .

This PR lets you use the 'Swap focused window with main window' hotkey to swap the two windows visible in the two-pane layout. 

This already works in any layout when only two windows are present on a screen. The two-pane layout is a bit different here as there are always only two windows visible even though more might be open.

### DIFF
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -670,11 +670,8 @@ extension WindowManager: WindowTransitionTarget {
         return windows.isWindowFloating(window)
     }
 
-    func getCurrentLayout() -> Layout<Application.Window>? {
-        guard let screenManager: ScreenManager<WindowManager<Application>> = focusedScreenManager(), let layout = screenManager.currentLayout else {
-            return nil
-        }
-        return layout
+    func currentLayout() -> Layout<Application.Window>? {
+        return focusedScreenManager()?.currentLayout
     }
 
     func activeWindows(on screen: Screen) -> [Window] {
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -670,6 +670,13 @@ extension WindowManager: WindowTransitionTarget {
         return windows.isWindowFloating(window)
     }
 
+    func getCurrentLayout() -> Layout<Application.Window>? {
+        guard let screenManager: ScreenManager<WindowManager<Application>> = focusedScreenManager(), let layout = screenManager.currentLayout else {
+            return nil
+        }
+        return layout
+    }
+
     func activeWindows(on screen: Screen) -> [Window] {
         return windows.activeWindows(onScreen: screen).filter { window in
             return window.shouldBeManaged() && !self.windows.isWindowFloating(window)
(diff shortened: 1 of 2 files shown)


Amethyst/Managers/WindowTransitionCoordinator.swift
```diff
@@ -26,6 +26,7 @@ protocol WindowTransitionTarget: class {
     func executeTransition(_ transition: WindowTransition<Window>)
 
     func isWindowFloating(_ window: Window) -> Bool
+    func getCurrentLayout() -> Layout<Application.Window>?
```
ianyh: It is a bit more idiomatic in swift to avoid the verb.
```swift
func currentLayout()
```

Amethyst/Managers/WindowTransitionCoordinator.swift
```diff
@@ -49,13 +50,16 @@ class WindowTransitionCoordinator<Target: WindowTransitionTarget> {
             return
         }
 
-        // if there are 2 windows, we can always swap.  Just make sure we don't swap focusedWindow with itself.
-        switch windows.count {
-        case 1:
+        if windows.count == 1 {
             return
-        case 2:
+        } else if windows.count == 2 || target?.getCurrentLayout()?.layoutKey == TwoPaneLayout<Window>.layoutKey {
```
ianyh: You can actually keep this as a `switch`. You can separate clauses in the case by commas.
```swift
case 2, /* layout key check */:
```

Amethyst/Managers/WindowManager.swift
```diff
@@ -670,6 +670,13 @@ extension WindowManager: WindowTransitionTarget {
         return windows.isWindowFloating(window)
     }
 
+    func getCurrentLayout() -> Layout<Application.Window>? {
+        guard let screenManager: ScreenManager<WindowManager<Application>> = focusedScreenManager(), let layout = screenManager.currentLayout else {
```
ianyh: Optional chaining might simplify this since we're just returning the value.

ianyh (changes requested):
Minor style comments.

Amethyst/Managers/WindowTransitionCoordinator.swift
rickythefox: Not sure about this. The switch is on an int value and the layout check returns a bool so this does not compile:
```case 2, target?.getCurrentLayout()?.layoutKey == TwoPaneLayout<Window>.layoutKey:```. Also the if statement here seems a bit more clear as we no longer are dependant only on the window count.

Amethyst/Managers/WindowManager.swift
rickythefox: Thanks, fixed.

Amethyst/Managers/WindowTransitionCoordinator.swift
rickythefox: Thanks, fixed.

Amethyst/Managers/WindowTransitionCoordinator.swift
ianyh: Fair point.

ianyh (approved):

PR #1242, opened 2022-05-16
Make main window focus behaviour more consistent

rickythefox:
* Focusing window 2+ in two-pane layout and invoking the shortcut caused a crash! (sorry)
* It's not completely clear what the shortcut should do depending on selection in two-pane layout.

I think this should simplify things - when the main window is focused and you invoke the "Swap with main window" shortcut it swaps the two top-most windows in the `windows` array. This makes the shortcut a convenient way to switch back and forth between two windows.

### DIFF
--- a/Amethyst/Managers/WindowTransitionCoordinator.swift
+++ b/Amethyst/Managers/WindowTransitionCoordinator.swift
@@ -52,12 +52,12 @@ class WindowTransitionCoordinator<Target: WindowTransitionTarget> {
 
         if windows.count == 1 {
             return
-        } else if windows.count == 2 || target?.currentLayout()?.layoutKey == TwoPaneLayout<Window>.layoutKey {
-            // Swap the two visible windows, keep focus on the main window if it already was there
-            target?.executeTransition(.switchWindows(focusedWindow, windows[1 - focusedIndex]))
-            if focusedWindow == windows[0] {
-                windows[1 - focusedIndex].focus()
-            }
+        }
+
+        if focusedIndex == 0 {
+            // If main window is focused - swap the two top-most windows, keep focus on the main window
+            target?.executeTransition(.switchWindows(focusedWindow, windows[1]))
+            windows[1].focus()
         } else {
             // Swap focused window with main window
             target?.executeTransition(.switchWindows(focusedWindow, windows[0]))
--- a/Amethyst/Managers/WindowTransitionCoordinator.swift
+++ b/Amethyst/Managers/WindowTransitionCoordinator.swift
@@ -50,7 +50,7 @@ class WindowTransitionCoordinator<Target: WindowTransitionTarget> {
             return
         }
 
-        if windows.count == 1 {
+        if windows.count <= 1 {
             return
         }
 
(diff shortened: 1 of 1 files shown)


Amethyst/Managers/WindowTransitionCoordinator.swift
```diff
@@ -52,12 +52,12 @@ class WindowTransitionCoordinator<Target: WindowTransitionTarget> {
...
-        } else if windows.count == 2 || target?.currentLayout()?.layoutKey == TwoPaneLayout<Window>.layoutKey {
-            // Swap the two visible windows, keep focus on the main window if it already was there
-            target?.executeTransition(.switchWindows(focusedWindow, windows[1 - focusedIndex]))
-            if focusedWindow == windows[0] {
-                windows[1 - focusedIndex].focus()
-            }
+        }
+
+        if focusedIndex == 0 {
+            // If main window is focused - swap the two top-most windows, keep focus on the main window
```
ianyh: Can you still restrict this to two-pane? I think it would be confusing for other layouts.

ianyh (changes requested):

Amethyst/Managers/WindowTransitionCoordinator.swift
rickythefox: Of course, but what should this shortcut do for other layouts if the main pane is already focused? 

* My intuition is it should swap the main window with the one that previously was the main window. This gives you a primary/secondary window to work with while saving one keypress. Might not be applicable to every layout, but maybe two-pane **and tall**?
* XMonad does nothing at all but it seems a waste of a perfectly good shortcut.

ianyh: I think for this change we should leave it as nothing, but I think it's worth doing the first change as a separate PR.

rickythefox: Pushed the fix

Amethyst/Managers/WindowTransitionCoordinator.swift
```diff
@@ -52,14 +52,17 @@ class WindowTransitionCoordinator<Target: WindowTransitionTarget> {
...
-            // Swap the two visible windows, keep focus on the main window if it already was there
-            target?.executeTransition(.switchWindows(focusedWindow, windows[1 - focusedIndex]))
-            if focusedWindow == windows[0] {
-                windows[1 - focusedIndex].focus()
-            }
-        } else {
-            // Swap focused window with main window
+        }
+
+        if focusedIndex == 0 && target?.currentLayout()?.layoutKey == TwoPaneLayout<Window>.layoutKey {
```
ianyh, line 57: We check for `windows.count == 1`, but we could theoretically have no windows, so I think we need to make sure that is not the case before we check index `1`. Or maybe change the check on the windows count above to be `<= 1`.

ianyh (changes requested):

Amethyst/Managers/WindowTransitionCoordinator.swift
rickythefox: Thanks, missed that.

ianyh (approved):

PR #1249, opened 2022-05-25
Fix ghost windows which have nil Space value under BSP Layout 

### LINKED ISSUE #1246: Tilling error under BSP Layout when typing with input method (like Ghost window)
**Describe the bug**
Under BSP layout, if type something with a system Input method (Chinese input method in my case), a minimal Input method words candidate window occur that make window management exception, as if there is a "ghost window" that will not go away

**Applications:**
What applications are involved?
- Input Method (Chinese IM. Japanese, Korean IM is the same I believe)

**To Reproduce** 
1. Open two Safari windows
2. Change Amethyst to BSP Layout
3. Input something at a Safari window
4. If this has't reproduce the bug, Change input method to English and change back and typed a few times more.

**Expected behavior**
- The upper window was reduced to half of the original size, like this has a 'Ghost window'
- Even when inputing is over, the IM window has disappeared, the 'Ghost window' will not go away

**Screenshots**

Screen Shot 2022-05-25 at 09 13 19

**Versions:**
 - macOS: 12.4
 - Amethyst: 0.15.6(93)

**Additional context**
- I suspect these issues are also related to this bug
-  #1054 
- #1169

LudwigWS:
Fix ghost windows which have nil Space value under BSP Layout (e.g. typing with Input Method).

A event about a window instance with a nil 'Space' property should be reported as unknown.

Fix #1246

### DIFF
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -334,12 +334,7 @@ final class WindowManager<Application: ApplicationType>: NSObject, Codable {
         guard let screen = window.screen() else {
             return
         }
-        let windowIDsArray = CGWindowsInfo.windowIDsArray(window)
-        guard let spaces = CGSCopySpacesForWindows(CGSMainConnectionID(), kCGSAllSpacesMask, windowIDsArray)?.takeRetainedValue() else {
-            return
-        }
-
-        let space = (spaces as NSArray as? [NSNumber])?.first?.intValue
+        let space = CGWindowsInfo.windowSpace(window)
 
         let windowChange: Change = windows.isWindowFloating(window) || space == nil ? .unknown : .add(window: window)
         markScreen(screen, forReflowWithChange: windowChange)
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -334,7 +334,7 @@ final class WindowManager<Application: ApplicationType>: NSObject, Codable {
         guard let screen = window.screen() else {
             return
         }
-        let windowIDsArray = [NSNumber(value: window.cgID() as UInt32)] as NSArray
+        let windowIDsArray = CGWindowsInfo.windowIDsArray(window)
         guard let spaces = CGSCopySpacesForWindows(CGSMainConnectionID(), kCGSAllSpacesMask, windowIDsArray)?.takeRetainedValue() else {
             return
         }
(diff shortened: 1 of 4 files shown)


Amethyst/Managers/WindowManager.swift
```diff
@@ -334,8 +334,14 @@ final class WindowManager<Application: ApplicationType>: NSObject, Codable {
         guard let screen = window.screen() else {
             return
         }
+        let windowIDsArray = [NSNumber(value: window.cgID() as UInt32)] as NSArray
```
ianyh: Could you make this a function on `CGWindowsInfo` or `CGSpacesInfo`? The manager is generic to aid in further testing, which requires isolating accesses to application specific functions like the CGS copy functions.

ianyh (changes requested):

Amethyst/Managers/WindowManager.swift
LudwigWS: > Could you make this a function on `CGWindowsInfo` or `CGSpacesInfo`? The manager is generic to aid in further testing, which requires isolating accesses to application specific functions like the CGS copy functions.

Of course, I'll do it later, I'm preparing my exam😭.

ianyh: No worries. Good luck with your exam!

Amethyst/Managers/WindowManager.swift
```diff
@@ -334,8 +334,14 @@ final class WindowManager<Application: ApplicationType>: NSObject, Codable {
         guard let screen = window.screen() else {
             return
         }
+        let windowIDsArray = CGWindowsInfo.windowIDsArray(window)
+        guard let spaces = CGSCopySpacesForWindows(CGSMainConnectionID(), kCGSAllSpacesMask, windowIDsArray)?.takeRetainedValue() else {
+            return
+        }
+
+        let space = (spaces as NSArray as? [NSNumber])?.first?.intValue
```
ianyh: Sorry, I meant that you should take this entire chunk and make it a function in CGInfo.

ianyh (changes requested):

Amethyst/Managers/WindowManager.swift
LudwigWS: > Sorry, I meant that you should take this entire chunk and make it a function in CGInfo.

Ok, I've been using a block application to control my computer, I only have one hour each day, hope I will do it correctly this night.😆

Amethyst/Managers/Windows.swift
```diff
@@ -41,13 +41,7 @@ extension WindowManager {
             }
 
             let screenWindows = windows.filter { window in
-                let windowIDsArray = [NSNumber(value: window.cgID() as UInt32)] as NSArray
```
ianyh, line 44: Nice, thanks for catching this, too.

ianyh (approved):

PR #1251, opened 2022-05-29
Return of the configuration file

ianyh:
- [x] Support yaml
- [x] Update tests to cover yaml configs in addition to json
- [x] Add example yaml config
- [x] Add config documentation


rickythefox:
image
We probably should be able to add window title restrictions?

joech4n:
> We probably should be able to add window title restrictions?

Is this currently possible? I don't see the option at https://github.com/ianyh/Amethyst/blob/development/docs/configuration-files.md

Edit: It also seems like any window title restriction configured via the GUI is wiped on restart when using `.amethyst.yml`. Perhaps this is better served by creating a new issue. Created https://github.com/ianyh/Amethyst/issues/1334.

PR #1276, opened 2022-07-02
Enable or disable AXEnhancedUserInterface as necessary

### LINKED ISSUE #1257: Occasional sluggish layout re-arrangement with Chrome
**Describe the bug**

I don't think that a description will do this justice, so I'll show a screen recording.

https://user-images.githubusercontent.com/8128941/173117608-901ece3c-bffd-455e-94a1-0156cf3bc4a9.mp4

You'll see that, when switching between layouts, the transitions are not instant, and I have to hit the layout key multiple times for things to look "correct".

This sluggishness also seems to happen when moving windows around inside of a layout:

https://user-images.githubusercontent.com/8128941/173117931-1375d224-dbfa-4cd1-a009-bb1732d83c5f.mp4

Typically, when I start Amethyst for the day, the animations for things like this are instant. But at some point (still trying to figure out what causes it) this sluggishness happens and it's pretty frustrating. Sometimes relaunching Amethyst fixes it, sometimes it doesn't. Often it can be "fixed" for a short amount of time if I reboot my machine, but it seems to always come back.

To make this even stranger, this has only been happening for a couple of weeks, and I've been using Amethyst for a few months with no issues. There has not been an update to Amethyst in this time that I can blame this behavior on. The only thing I can think of is that I tried Yabai for a short period of time last week (disabled Amethyst during that time) but I have since gotten rid of that and got back to Amethyst. I unfortunately don't remember if this behavior started before or after I installed Yabai. It feels weird to place the blame on that though, considering that it's been removed from my system, but it's 

**Applications:**

> What applications are involved?

This seems to happen for every application, not just the app that I used in the screen recording above.

**To Reproduce**

I'm still not sure how I get to this state, but I'll update this message if I have a reliable setup.

**Expected behavior**

Switching between layouts, moving windows around, etc. should be instant, and shouldn't require executing the same command multiple times.

**Versions:**
 - macOS: 12.4 on an M1 Pro MBP
 - Amethyst: 0.15.6
 
**Debug Info**
```
$ /Applications/Amethyst.app/Contents/MacOS/Amethyst --debug-info
Version: 0.15.6 (93)

OS version: Version 12.4 (Build 21F79)

Screens:
	(0.0, 0.0, 1728.0, 1117.0) [(0.0, 0.0, 1728.0, 1117.0)]
	(1728.0, 0.0, 2560.0, 1440.0) [(1728.0, -323.0, 2560.0, 1440.0)]
	(4288.0, 0.0, 2560.0, 1440.0) [(4288.0, -323.0, 2560.0, 1440.0)]

Configuration:
window-minimum-width: 0
floating-is-blacklist: 1
window-margins: 1
mouse-resizes-windows: 1
enables-layout-hud: 1
debug-layout-info: 0
smart-window-margins: 1
window-resize-step: 5
layouts: (
    "two-pane",
    tall,
    fullscreen
)
window-margin-size: 10
mod1: (
    option,
    shift
)
restore-layouts-on-launch: 0
screen-padding-bottom: 10
focus-follows-mouse: 0
follow-space-thrown-windows: 1
new-windows-to-main: 0
enables-layout-hud-on-space-change: 1
use-canary-build: 1
ignore-menu-bar: 0
screen-padding-top: 10
window-minimum-height: 0
mouse-follows-focus: 1
screen-padding-right: 10
mouse-swaps-windows: 1
float-small-windows: 1
mod2: (
    option,
    shift,
    control
)
floating: (
        {
        id = "com.apple.systempreferences";
        "window-titles" =         (
        );
    },
        {
        id = "io.appmakes.Paletro";
        "window-titles" =         (
        );
    },
        {
        id = "com.zsa.wally";
        "window-titles" =         (
        );
    },
        {
        id = "fm.qmk.toolbox";
        "window-titles" =         (
        );
    },
        {
        id = "com.hogbaysoftware.Bike";
        "window-titles" =         (
        );
    }
)
screen-padding-left: 10
```

ianyh:
Should fix #1257

### DIFF
--- a/Amethyst/Layout/ReflowOperation.swift
+++ b/Amethyst/Layout/ReflowOperation.swift
@@ -217,7 +217,9 @@ struct FrameAssignment<Window: WindowType> {
             // Just resize the window first to see what the dimensions end up being
             // Sometimes applications have internal window requirements that are not exposed to us directly
             finalFrame.origin = window.frame().origin
-            window.setFrame(finalFrame, withThreshold: CGSize(width: 1, height: 1))
+            DispatchQueue.main.sync {
+                window.setFrame(finalFrame, withThreshold: CGSize(width: 1, height: 1))
+            }
 
             // With the real height we can update the frame to account for the current size
             finalFrame.size = CGSize(
@@ -230,6 +232,8 @@ struct FrameAssignment<Window: WindowType> {
 
         // Move the window to its final frame
         finalFrame.origin = finalOrigin
-        window.setFrame(finalFrame, withThreshold: CGSize(width: 1, height: 1))
+        DispatchQueue.main.sync {
+            window.setFrame(finalFrame, withThreshold: CGSize(width: 1, height: 1))
+        }
     }
 }
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -647,7 +647,9 @@ extension WindowManager: WindowTransitionTarget {
             window.move(toSpace: targetSpace.id)
             // necessary to set frame here as window is expected to be at origin relative to targe screen when moved, can be improved.
             let newFrame = targetScreen.frameWithoutDockOrMenu()
-            window.setFrame(newFrame, withThreshold: CGSize(width: 25, height: 25))
+            DispatchQueue.main.sync {
+                window.setFrame(newFrame, withThreshold: CGSize(width: 25, height: 25))
+            }
             markScreen(targetScreen, forReflowWithChange: .add(window: window))
             if UserConfiguration.shared.followWindowsThrownBetweenSpaces() {
                 window.focus()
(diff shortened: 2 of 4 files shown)


nzhl:
I compiled the app from your branch, but it seems that the issue still exists (try puting two chrome windows into a space and doing window rearrange).

ianyh:
My chrome never does this, but I can maybe manually apply the value to see if I can make it start doing it.

ianyh:
This actually still doesn't work because of interleaving of setting frames.

ianyh:
Seems to actually work now, though I have to think about the serialization of setting frames.

ianyh:
I tried a lot of different approaches, but I don't think I can get the interleaving to behave correctly without implementing some kind of lock on elements that seems a little overboard for most use cases.

PR #1284, opened 2022-07-20
add "Terminologies"

thai-d-v:
I re-format README and added "Terminologies" to explain how Amethyst works more clearly. You can see more detail in commit messages.


ianyh (changes requested):
I think it's generally a good idea to have a terminologies reference, but could you take this out into a page in docs?

thai-d-v:
I hope that is what you meant.

ianyh:
That is what I meant, yes. Thanks.

PR #1293, opened 2022-08-16
Added option to disable screen padding on built-in display

anthdono:
Screen padding is often used on external displays to better position content in the line of sight, however the effects are not desirable on a laptop screen.

This option adds the ability to disable screen padding for the in-built display, whilst still having it enabled on external monitors. The checkbox option is disabled by default. 

The feature is useful when not using clamshell mode, ie when the built-in display is being used alongside external displays. Likewise, when working solely from the laptop, this feature eliminates the need to manually decrement the padding that was configured for external monitors.

general-preferences

### DIFF
--- a/Amethyst/Model/Screen.swift
+++ b/Amethyst/Model/Screen.swift
@@ -100,6 +100,12 @@ struct AMScreen: ScreenType {
             frame.size.height = windowMinimumHeight
         }
 
+         var isDisablePaddingOnBuiltinDisplay: Bool =
+             UserConfiguration.shared.disablePaddingOnBuiltinDisplay()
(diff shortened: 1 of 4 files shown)


ianyh:
Cool. Can you add this to the config documentation?

anthdono:
Have added feature to the config documentation and rebased my fork with latest upstream developer branch [3c2835a]. Cheers!

ianyh (approved):

ianyh:
Thanks!

Mathemmagician:
Is there a way to disable padding on external only? I know it's an edge case, but trying to deal with some broken pixels on built-in monitor, and would really love to exclude them

PR #1307, opened 2022-09-09
adds rest of layouts to sample config file

dmartzol:
When creating a config file for the first time, I tried to add the Two Pane layout with the keys `twopane`, `two_pane`, `twoPane` before cloning the repo and using grep to find out that the correct key was `two-pane`.

Having all layouts by default might save time for people who don't know what the proper consensus is for yaml files and such keys :)


ianyh:
Thanks!

PR #1322, opened 2022-10-13
Send console.log in custom layouts to syslog

### LINKED ISSUE #1294: javascript console log for custom layouts
It would be helpful to have a way to see the javascript console log  when developing custom layouts.

One possibility would be that console.log() content gets written to "syslog", and users can view this with the Mac OS Console app.

Another solution could be that the app opens a window/dialog which does nothing other than display the messages that layouts write using "console.log". 

**Additional context**
This is useful for someone developing or troubleshooting a custom layout.

rickythefox:
Hook console.log in custom layouts to write to syslog.

Closes #1294

### DIFF
--- a/Amethyst/Layout/Layouts/CustomLayout.swift
+++ b/Amethyst/Layout/Layouts/CustomLayout.swift
@@ -49,6 +49,12 @@ class CustomLayout<Window: WindowType>: StatefulLayout<Window> {
             return nil
         }
 
+        context.evaluateScript("var console = { log: function(message) { _consoleLog(message) } }")
+        let consoleLog: @convention(block) (String) -> Void = { message in
+            log.debug(message)
+        }
+        context.setObject(unsafeBitCast(consoleLog, to: AnyObject.self), forKeyedSubscript: "_consoleLog" as (NSCopying & NSObjectProtocol))
+
         do {
             context.evaluateScript(try String(contentsOf: self.fileURL))
         } catch {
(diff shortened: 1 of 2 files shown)


Amethyst/Layout/Layouts/CustomLayout.swift
```diff
@@ -49,6 +49,12 @@ class CustomLayout<Window: WindowType>: StatefulLayout<Window> {
             return nil
         }
 
+        context.evaluateScript("var console = { log: function(message) { _consoleLog(message) } }")
+        let consoleLog: @convention(block) (String) -> Void = { message in
+            NSLog(message)
```
ianyh: Can you use the logging used elsewhere rather than directly to syslog?

ianyh (changes requested):

Amethyst/Layout/Layouts/CustomLayout.swift
rickythefox:
```    
func applicationDidFinishLaunching(_ notification: Notification) {
#if DEBUG
    log.addDestination(ConsoleDestination())
#endif

log.info("Logging is enabled")
```
This tells me that logging is only enabled for debug builds, right? I don't think using console.log to debug a custom layout should require a debug build of Amethyst?

ianyh: There used to be a command-line option for sending logs to stdout. Looks like I lost it somewhere along the way.

rickythefox: @ianyh added `--log` command line option and switched to the logging framework used by you.

ianyh (approved):

PR #1324, opened 2022-10-18
Support XDG config

### LINKED ISSUE #1268: Support XDG Base Directory Specification for `amethyst.yml`
First of all, thanks for providing a configuration file, I really appreciate to be able to check in my configuration in a source control system.

# Background
Lately the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) has been gaining quite a steady [popularity](https://wiki.archlinux.org/title/XDG_Base_Directory#Support), even though it's been around for a long time already. The basic idea is to standardize where programs put their files e.g. all config files in a certain directory. The benefits for end users are many, including:

* No polluted $HOME/
* Can easily grep for only config files in `$XDG_CONFIG_HOME`
* Can decide where to store configurations in more complex setups (multi system setup e.g.)
* Easily wipe state by removing `$XDG_STATE_HOME` or even `$XDG_DATA_HOME`, but without destroying and configuration files.

The first point is maybe what most end users enjoy the most though.

# How
Specifically for this project, this should mean that the search for the configuration file should be done like:

1. **If** `$XDG_CONFIG_HOME` is *set*, look for `$XDG_CONFIG_HOME/amethyst/amethyst.yml`
1. **elseIf** `$XDG_CONFIG_HOME` is *unset*, use the default value for the envvar which is `$HOME/.config`. Thus, the config file to look for is at `$HOME/.config/amethyst/amethyst.yml`
1. **else** fall-back to current legacy `$HOME/.amethyst.yml`

The first two steps to can be codifed in bash like:
 `${XDG_CONFIG_HOME:-$HOME/.config}/amethyst/amethyst.yml`

cowboy-bebug:
Thanks for the configuration feature 🙏

Here's the enhancement to support `$XDG_CONFIG_HOME` (or just `$HOME/.config`).

Closes #1268.

### DIFF
--- a/Amethyst/Preferences/UserConfiguration.swift
+++ b/Amethyst/Preferences/UserConfiguration.swift
@@ -319,12 +319,15 @@ class UserConfiguration: NSObject {
     }
 
     private func loadConfigurationFile() {
+        let xdgConfigPath = ProcessInfo.processInfo.environment["XDG_CONFIG_HOME"] ?? NSHomeDirectory().appending("/.config")
+        let amethystXDGConfigPath = xdgConfigPath.appending("/amethyst")
         let amethystYAMLConfigPath = NSHomeDirectory().appending("/.amethyst.yml")
         let amethystJSONConfigPath = NSHomeDirectory().appending("/.amethyst")
         let defaultAmethystConfigPath = Bundle.main.path(forResource: "default", ofType: "amethyst")
 
-        if FileManager.default.fileExists(atPath: amethystYAMLConfigPath, isDirectory: nil) {
-            configurationYAML = yamlForConfig(at: amethystYAMLConfigPath)
+        var isDirectory: ObjCBool = false
+        if FileManager.default.fileExists(atPath: amethystXDGConfigPath, isDirectory: &isDirectory) {
+            configurationYAML = yamlForConfig(at: isDirectory.boolValue ? amethystXDGConfigPath.appending("/amethyst.yml") : amethystYAMLConfigPath)
 
             if configurationYAML == nil {
                 log.error("error loading configuration as yaml")
@@ -334,8 +337,8 @@ class UserConfiguration: NSObject {
                 alert.messageText = "Error loading configuration"
                 alert.runModal()
             }
-        } else if FileManager.default.fileExists(atPath: amethystJSONConfigPath, isDirectory: nil) {
-            configurationJSON = jsonForConfig(at: amethystJSONConfigPath)
+        } else if FileManager.default.fileExists(atPath: amethystXDGConfigPath, isDirectory: &isDirectory) {
+            configurationJSON = jsonForConfig(at: isDirectory.boolValue ? amethystXDGConfigPath.appending("/amethyst") : amethystJSONConfigPath)
 
             if configurationJSON == nil {
                 log.error("error loading configuration as json")


DitchComfort:
Does this mean I can now use a location like:
$HOME/.config/amethyst/amethyst.yml
or
${XDG_CONFIG_HOME:-$HOME/.config}/amethyst/amethyst.yml

Amethyst/Preferences/UserConfiguration.swift
```diff
@@ -334,8 +337,8 @@ class UserConfiguration: NSObject {
                 alert.messageText = "Error loading configuration"
                 alert.runModal()
             }
-        } else if FileManager.default.fileExists(atPath: amethystJSONConfigPath, isDirectory: nil) {
-            configurationJSON = jsonForConfig(at: amethystJSONConfigPath)
+        } else if FileManager.default.fileExists(atPath: amethystXDGConfigPath, isDirectory: &isDirectory) {
+            configurationJSON = jsonForConfig(at: isDirectory.boolValue ? amethystXDGConfigPath.appending("/amethyst") : amethystJSONConfigPath)
```
ianyh, line 341: Should it be `/.amethyst` rather than `/amethyst`?

ianyh (changes requested):

Amethyst/Preferences/UserConfiguration.swift
DitchComfort: If its in your home directory it usually is .amethyst.yml, but if you follow the XDG, it should be $HOME/.config/amethyst/amethyst.yml

DitchComfort: Home directory: $HOME/.amethyst.yml

XDG-based: $HOME/.config/amethyst/amethyst.yml

cowboy-bebug: Yeah that's how they usually are because files don't need to be hidden if it's not in the home directory.

Some examples on my machine:
```
~/.config/alacritty/alacritty
~/.config/nvim/init.lua
~/.config/git/config
```

One exception I see is [zsh](https://zsh.sourceforge.io/Doc/Release/Files.html) where the hidden file notation is consistently used (but they use `$ZDOTDIR` rather than `$XDG_CONFIG_HOME`):
```bash
> ls -la .config/zsh/
total 120
drwxr-xr-x  6 eric  staff    192 18 Oct 11:01 .
drwxr-xr-x  3 eric  staff     96 12 Oct 11:07 ..
-rw-r--r--  1 eric  staff  47799 12 Oct 11:12 .zcompdump
-rwxr-xr-x@ 1 eric  staff    642 18 Oct 20:40 .zlogin
-rw-r--r--  1 eric  staff     83 18 Oct 10:24 .zprofile
-rw-r--r--  1 eric  staff   1095 18 Oct 11:01 .zshrc
```

@ianyh What do you prefer?

ianyh: Got it. The intention is to meet a standard, and it seems like there's at least a convention, so let's go with not using hidden file name.

ianyh (approved):
Nice change! Thanks.

DitchComfort:
Can I actually use ${XDG_CONFIG_HOME:-$HOME/.config}/amethyst/amethyst.yml now?

ianyh:
I will need to make a new release first.

DitchComfort:
Okey super nice. And what about Ventura, ready?

Issue #1321, opened 2022-10-12
Update .amethyst.default.yml to have all defaults as example


wley3337:
**Is your feature request related to a problem? Please describe.**
I'm a bit lost in what the options are and how to describe them in the current `.amethyst.sample.yml` file

**Describe the solution you'd like**
Add or replace the `.amethyst.sample.yml` with one with all the defaults

Proposed change
```yml
# Default settings for amethyst
# Repo: `https://github.com/ianyh/Amethyst`
#
# Move this file to: `~/.amethyst.yml`
# In order to register changes restart Amethyst.
# If you experience issues pulling in the changes you can also quit Amethyst and run: `rm ~/Library/Preferences/com.amethyst.Amethyst.plist`
# This removes the current preferences and causes a re-load of this file when amethyst restarts.

# layouts	Ordered list of layouts to use by layout key (default tall, wide, fullscreen, and column).
layouts:
  - tall
  - fullscreen
  # - tall-right
  - wide
  # - two-pane
  # - middle-wide
  # - 3column-left
  # - 3column-middle
  # - 3column-right
  # - 4column-left
  # - 4column-right
  - column
  # - row
  # - floating
  # - widescreen-tall
  # - widescreen-tall-right
  # - bsp

#	First mod (default option + shift).
mod1:
  - option
  - shift
  # - control
  # - command

# Second mod (default option + shift + control).
mod2:
  - option
  - shift
  - control
  # - command

# Commands:
# special key values
# "space"
# "enter"
# "up"
# "right"
# "down"
# "left"

# Move to the next layout in the list.
cycle-layout:
  mod: mod1
  key: "space"

# Move to the previous layout in the list.
cycle-layout-backward:
  mod: mod2
  key: "space"

# Shrink the main pane by a percentage of the screen dimension as defined by window-resize-step. Note that not all layouts respond to this command.
shrink-main:
  mod: mod1
  key: h

# Expand the main pane by a percentage of the screen dimension as defined by window-resize-step. Note that not all layouts respond to this command.
expand-main:
  mod: mod1
  key: l

# Increase the number of windows in the main pane. Note that not all layouts respond to this command.
increase-main:
  mod: mod1
  key: ","

# Decrease the number of windows in the main pane. Note that not all layouts respond to this command.
decrease-main:
  mod: mod1
  key: "."

# General purpose command for custom layouts. Functionality is layout-dependent.
# command1:
#   mod: <NONE>
#   key: <NONE>
[...]

PR #1326, opened 2022-10-19
Issue 1321: update amethyst.sample.yml include all defaults

wley3337:
Update `amethyst.sample.yml` to include all the defaults and the text description of what they do

[Issue 1321](https://github.com/ianyh/Amethyst/issues/1321)

### DIFF
--- a/.amethyst.sample.yml
+++ b/.amethyst.sample.yml
@@ -148,7 +148,7 @@ swap-cw:
 # Swap the position of the currently focused window with the main window in the list.
 swap-main:
   mod: mod1
-  key: "enter"
+  key: enter
 
 # Move focus to the n-th screen in the list; e.g., focus-screen-3 will move mouse focus to the 3rd screen. Note that the main window in the given screen will be focused.
 #focus-screen-n:
--- a/.amethyst.sample.yml
+++ b/.amethyst.sample.yml
@@ -1,42 +1,280 @@
-# Use tall and fullscreen layouts
+# Default settings for amethyst
+# Repo: `https://github.com/ianyh/Amethyst`
+#
+# Move this file to: `~/.amethyst.yml`
+# In order to register changes restart Amethyst.
+# If you experience issues pulling in the changes you can also quit Amethyst and run: `rm ~/Library/Preferences/com.amethyst.Amethyst.plist`
+# This removes the current preferences and causes a re-load of this file when amethyst restarts.
+
+# layouts	Ordered list of layouts to use by layout key (default tall, wide, fullscreen, and column).
 layouts:
-    - tall
-    - fullscreen
-    - tall-right
-    - wide
-    - two-pane
-    - 3column-left
-    - middle-wide
-    - 3column-right
-    - 4column-left
-    - 4column-right
-    - column
-    - row
-    - floating
-    - widescreen-tall
-    - widescreen-tall-right
-    - bsp
+  - tall
+  - fullscreen
+  # - tall-right
+  - wide
+  # - two-pane
+  # - middle-wide
+  # - 3column-left
+  # - 3column-middle
+  # - 3column-right
+  # - 4column-left
+  # - 4column-right
+  - column
+  # - row
+  # - floating
+  # - widescreen-tall
+  # - widescreen-tall-right
+  # - bsp
 
-# Add "command" to the default mods
+#	First mod (default option + shift).
 mod1:
-    - option
-    - shift
-    - command
+  - option
+  - shift
+  # - control
+  # - command
+
+# Second mod (default option + shift + control).
 mod2:
-    - option
-    - shift
-    - control
-    - command
+  - option
+  - shift
+  - control
+  # - command
 
-# Modify the amethyst relaunch to use x and not z
+# Commands:
+# special key values
+# "space"
+# "enter"
+# "up"
+# "right"
+# "down"
+# "left"
+
+# Move to the next layout in the list.
+cycle-layout:
+  mod: mod1
+  key: "space"
+
+# Move to the previous layout in the list.
+cycle-layout-backward:
+  mod: mod2
+  key: "space"
+
+# Shrink the main pane by a percentage of the screen dimension as defined by window-resize-step. Note that not all layouts respond to this command.
+shrink-main:
+  mod: mod1
+  key: h
+
+# Expand the main pane by a percentage of the screen dimension as defined by window-resize-step. Note that not all layouts respond to this command.
+expand-main:
+  mod: mod1
+  key: l
+
+# Increase the number of windows in the main pane. Note that not all layouts respond to this command.
+increase-main:
+  mod: mod1
+  key: ","
+
+# Decrease the number of windows in the main pane. Note that not all layouts respond to this command.
+decrease-main:
+  mod: mod1
+  key: "."
+
+# General purpose command for custom layouts. Functionality is layout-dependent.
+# command1:
+#   mod: <NONE>
+#   key: <NONE>
+# General purpose command for custom layouts. Functionality is layout-dependent.
+# command2:
+#   mod: <NONE>
+#   key: <NONE>
+# General purpose command for custom layouts. Functionality is layout-dependent.
+# command3:
+#   mod: <NONE>
+#   key: <NONE>
+# General purpose command for custom layouts. Functionality is layout-dependent.
+# command4:
+#   mod: <NONE>
+#   key: <NONE>
+
+# Focus the next window in the list going counter-clockwise.
+focus-ccw:
+  mod: mod1
+  key: j
+
+# Focus the next window in the list going clockwise.
+focus-cw:
+  mod: mod1
+  key: k
+
+# Focus the main window in the list.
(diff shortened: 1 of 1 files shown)


.amethyst.sample.yml
```diff
@@ -1,42 +1,280 @@
-# Use tall and fullscreen layouts
+# Default settings for amethyst
+# Repo: `https://github.com/ianyh/Amethyst`
+#
+# Move this file to: `~/.amethyst.yml`
+# In order to register changes restart Amethyst.
+# If you experience issues pulling in the changes you can also quit Amethyst and run: `rm ~/Library/Preferences/com.amethyst.Amethyst.plist`
+# This removes the current preferences and causes a re-load of this file when amethyst restarts.
```
agoodshort: I don't know all the mechanics behind macOS and preferences, but I have been playing with preferences yesterday for thie issue #1348  and the `/Library/Preferences/com.amethyst.Amethyst.plist` file does not seem to always be generated (at least on my Mac mini macOS Ventura). However, you can reset your preferences using `defaults delete com.amethyst.Amethyst.plist`

```suggestion
# If you experience issues pulling in the changes you can also quit Amethyst and run: `defaults delete com.amethyst.Amethyst.plist`
# This removes the current preferences and causes Amethyst to restart with default preferences and pull configs from this file.
```

.amethyst.sample.yml
```diff
@@ -1,42 +1,280 @@
...
+  # - command
+
+# Commands:
+# special key values
+# "space"
+# "enter"
+# "up"
+# "right"
+# "down"
+# "left"
```
agoodshort: AFAIK YAML requires quotes only for the special characters below:
```
: { } [ ] , & * # ? | - < > = ! % @ `
```

Just a matter of style and preferences I would use the below unquoted.

I would add `,` `.` as these are special characters implemented by default. And I would use single quote as best practice. (i.e. `"\"` breaks the config, but `'\'` doesn't. However, a simple `key: \` works too, as it is not considered a special character...)

```suggestion
# space
# enter
# up
# right
# down
# left

# special characters require quotes
# '.'
# ','
```

.amethyst.sample.yml
```diff
@@ -1,42 +1,280 @@
...
+
+# Move to the next layout in the list.
+cycle-layout:
+  mod: mod1
+  key: "space"
```
agoodshort: Based on suggestion above
```suggestion
  key: space
```

.amethyst.sample.yml
```diff
@@ -1,42 +1,280 @@
...
+
+# Move to the previous layout in the list.
+cycle-layout-backward:
+  mod: mod2
+  key: "space"
```
agoodshort: Based on suggestion above
```suggestion
  key: space
```

.amethyst.sample.yml
```diff
@@ -1,42 +1,280 @@
...
+
+# Expand the main pane by a percentage of the screen dimension as defined by window-resize-step. Note that not all layouts respond to this command.
+expand-main:
+  mod: mod1
+  key: l
+
+# Increase the number of windows in the main pane. Note that not all layouts respond to this command.
+increase-main:
+  mod: mod1
+  key: ","
```
agoodshort: Based on suggestion above
```suggestion
  key: ','
```

.amethyst.sample.yml
```diff
@@ -1,42 +1,280 @@
...
+
+# Decrease the number of windows in the main pane. Note that not all layouts respond to this command.
+decrease-main:
+  mod: mod1
+  key: "."
```
agoodshort: Based on suggestion above
```suggestion
  key: '.'
```

.amethyst.sample.yml
```diff
@@ -1,42 +1,280 @@
...
+
+# Swap the position of the currently focused window with the next window in the list going clockwise.
+swap-cw:
+  mod: mod2
+  key: k
+
+# Swap the position of the currently focused window with the main window in the list.
+swap-main:
+  mod: mod1
+  key: "enter"
```
agoodshort: Based on suggestion above
```suggestion
  key: enter
```

.amethyst.sample.yml
```diff
@@ -1,42 +1,280 @@
...
+#   key: u
+# Move the currently focused window to the n-th space; e.g., throw-space-3 will move the window to the 3rd space.
+# throw-space-<screen-number>:
+#   mod: mod1
+#   key: i
+
+# Move the currently focused window to the space to the left.
+throw-space-left:
+  mod: mod2
+  key: "left"
```
agoodshort: Based on suggestion above
```suggestion
  key: left
```

.amethyst.sample.yml
```diff
@@ -1,42 +1,280 @@
...
+
+# Move currently the focused window to the space to the right.
+throw-space-right:
+  mod: mod2
+  key: "right"
```
agoodshort: Based on suggestion above
```suggestion
  key: right
```

agoodshort (changes requested):
Few comments/suggested changes to make this starter config file a bit more resilient

.amethyst.sample.yml
```diff
@@ -1,42 +1,280 @@
-# Use tall and fullscreen layouts
+# Default settings for amethyst
+# Repo: `https://github.com/ianyh/Amethyst`
+#
+# Move this file to: `~/.amethyst.yml`
+# In order to register changes restart Amethyst.
+# If you experience issues pulling in the changes you can also quit Amethyst and run: `rm ~/Library/Preferences/com.amethyst.Amethyst.plist`
+# This removes the current preferences and causes a re-load of this file when amethyst restarts.
+
+# layouts	Ordered list of layouts to use by layout key (default tall, wide, fullscreen, and column).
```
agoodshort: Tab confuses syntax highlighting
```suggestion
# layouts - Ordered list of layouts to use by layout key (default tall, wide, fullscreen, and column).
```

.amethyst.sample.yml
```diff
@@ -1,42 +1,280 @@
...
+  # - 4column-left
+  # - 4column-right
+  - column
+  # - row
+  # - floating
+  # - widescreen-tall
+  # - widescreen-tall-right
+  # - bsp
+
+#	First mod (default option + shift).
```
agoodshort: Tab confuses syntax highlighting
```suggestion
# First mod (default option + shift).
```

agoodshort (changes requested):
2 minor suggestion for better readability through an IDE

wley3337:
@goodshort Sorry for the delay and thank you so much for the review! I tested out with my local and the `"` around the commands are definitely not needed ( outside the special characters ). Accepted all the suggestions. It's odd though there was only a single space on my local version for the comments rather than a tab space. I wonder why that looked like an indent 🤷

Let me know if you'd like me to squash and push back up :)

agoodshort (dismissed):
Works fine on my machine. File is now a lot more readable and well documented. Thanks @wley3337

cowboy-bebug (commented):
Some defaults such as `select-*-layout` are missing from [default.amethyst](https://github.com/ianyh/Amethyst/blob/development/Amethyst/default.amethyst) 😃

jabkoo:
+1 for @cowboy-bebug's suggestion. `select-*-layout` commands are not documented anywhere, so I would also like to see it mentioned in docs/configuration-files.md.

wley3337:
Hi all! 

Been a bit swamped between work and life. Going to try and sort out the `select-*-layout` stuff in the next few days and get this updated

Indyandie:
@wley3337 please note that this bug (#1419) could affect some of the configs

wley3337:
Hi all!

Thank you all for your comments and patience. @Indyandie I'm not sure how to handle the possible conflicts. I added a block comment at the top to help with troubleshooting. Other than that, the only thing I can think of is to comment out the default values that are in `default.amethyst` and add the comment on each of those lines.

Something like:
```yml
# Commented out to avoid conflicts with `default.amethyst`
# only uncomment if you're going to change the default value below
# Move to the next layout in the list
# cycle-layout:
#   mod: mod1
#   key: space
```

ianyh:
I might want to make some tweaks to this. Is it possible to give me permissions to push to this branch?

vicmunoz:
Thanks!
Only missing:


window-max-count | The max number of windows that may be visible on a screen at one time before additional windows are minimized. A value of 0 disables the feature. (default 0)
-- | --


disable-screen-padding-on-inbuilt | true to disable screen padding on in-built display (default false).
-- | --

and a typo, is mod2 in default
swap-screen-cw:
  mod: mod2
  key: l

prettymuchbryce:
We should merge this. Even if it's incomplete it is already way better than the current sample file, and this PR has been open for nearly 6 months.

wley3337:
Added the missing two defaults above and fixed typo

wley3337:
Oh, @ianyh I __think__ I granted you access to push/pull let me know if that didn't work :) I pulled right before addressing the two comments with no diff so not sure if Github let you know :)

ianyh (approved):

PR #1330, opened 2022-10-24
Feature/two row layout

joelee:
This is my proposal for a two-row Layout, a version of *Wide*, with the Main Pane at the bottom Row (See the image below). 

This is particularly useful if you have a large 48" TV on your desk where the bottom 2/3 is in your direct line of sight (Main Pane), and the top 1/3 is in your peripheral vision (Secondary Pane).
The maximum number of windows in the Panes is limited to 3 for Main, and 4 for Secondary, but can be changed using the custom layout command.

image


ianyh:
Is there a reason to do this natively versus a custom layout?

ianyh:
Having thought about it, I do think this should be a custom layout and not native.

Issue #1258, opened 2022-06-13
MacOS Ventura: Per-"stage" layouts with "Stage manager"


lougreenwood:
**Is your feature request related to a problem? Please describe.**

MacOS Ventura introduces stage manager, at a glance it seems to work with Amethyst, however it would be much more powerful if the tiling layout was per "stage" (group of apps), so I could group together apps related to email on one "stage" and development on another.

Switching between these app groups would restore the tiled layout of those apps specific to the group

tim-hm:
Because tiling applies to all windows in the space window area is calculated assuming all windows are displayed. However, when stage manager presents a stage/subset of those windows the tiling applied doesn't make sense.

Perhaps there is no intention of support stages and if so then maybe the documentation can be updated? Otherwise, I think @lougreenwood's suggestion above would make for a wonderful UX.

ianyh:
There's intention, I just haven't really had time. Someone is welcome to make the contribution in the meantime.

tim-hm:
Sorry wasn't trying to be inconsiderate and I totally appreciate how valuable your time is! Fantastic to hear that its planned 🙏

Muzosh:
When this goes to development, **I personally would prefer to just "apply tiling just on displayed windows" (i.e. those that are in currently open 'stage') as I think it might be more simple and faster, but much needed change.** Currently I view it as a bug and this would fix this bug.

Let's assume I have 3 windows open with "Tall" layout. If I group two windows together in one stage and switch to the other stage, the third window with its own stage automatically takes 25% in right-bottom screen. I would assume, that if it's displayed alone (since other two are hidden in another stage), it would be displayed in fullscreen (the same way the "Tall" behaves when you have only one window opened).

Afterwards, **some time later with no rush, a per-stage tilling management might be welcomed**. I assume there might be some added layer of saved configuration, thus complicating development. Currently, I guess there needs to be information stored about which tiling layout is applied _to which space_. With this, we would need another layer of information about which tiling layout is applied _to each stage on each space_. I consider this as just a new feature, not a bug.

Currently, the OP's use case can be done with two spaces so I would prefer the development of this feature to go that "fast and simple" way first.
> so I could group together apps related to email on one "stage" and development on another.

Muzosh:
[...]

PR #1331, opened 2022-10-27
Fix Amethyst behavior with Stage Manager from MacOS Ventura 

Muzosh:
This PR should fix the Amethyst usage with Stage Manager, see https://github.com/ianyh/Amethyst/issues/1258#issuecomment-1293792227


ianyh (changes requested):
This could inhibit desired behavior involving minimized windows in general. For example, a window that is minimized before Amethyst starts that is then unminimized independent of staging could be lost.

Edit: Sorry if this review was premature.

Muzosh:
> This could inhibit desired behavior involving minimized windows in general. For example, a window that is minimized before Amethyst starts that is then unminimized independent of staging could be lost.
> 
> Edit: Sorry if this review was premature.

Well, I just tried it and it seems that my assumptions were wrong. It basically behaves the same way.
So maybe Stage Manager does not just minimize windows? I have tried to find some development documentation so we as developer can recognize, which windows are displayed and which ones are hidden by Stage Manager, but I did not find anything.

Muzosh:
From what I was able to read from the code, the `Window:shouldBeManaged()` method is called every time windows are reevaluated in `WindowsManager:reevaluateWindows()` - this function does not run repeatedly?

Muzosh:
So, a loooong over-due update. I have finally found time and reason to look at this again and it seems that there is no way of knowing, whether the app window is hidden in the stage manager or not.

Firstly, the only mention of Stage Manager in relation to AppKit is [here](https://developer.apple.com/videos/play/wwdc2022/10074/?time=45). As far as my understanding goes, they basically say: "Don't worry about Stage Manager, it will work as expected based on what `collectionBehavior` you choose". So this is basically AppKit->Stage Manager link, but we want to get information from StageManager to he AppKit.

I also found, that similar buried problem was discussed in [yabai](https://github.com/koekeishiya/yabai/issues?q=stage+manager), however, no solutions so far.

Issue #1306, opened 2022-09-09
Throw a window to other spaces will follow the window to that space even when the settings is unchecked.


h9nara:
**Describe the bug**
Throw a window to other spaces and the focus will follow the window to that space even when the setting 'Follow thrown windows between spaces' is unchecked in preferences.

**Applications:**
Amethyst

**To Reproduce**
- Uncheck 'Follow thrown windows between spaces' in preferences
- Throw a window to another space and focus will follow

**Expected behavior**
The focus should stay at the space from which the window is thrown.

**Screenshots**

**Versions:**
 - macOS: Appears both on 12.4 and 12.5.1
 - Amethyst: 0.16.0 (94)
 
**Debug Info**
```
$ /Applications/Amethyst.app/Contents/MacOS/Amethyst --debug-info [--include-apps]
```
Note: `--include-apps` will list your manageable applications, but is optional if you don't want to list that.

**Additional context**

mike-lloyd03:
I'm noticing this issue as well on the latest update of Amethyst and it wasn't happening before updating. I'm on MacOS 12.5.1.

I've tried checking the Follow option and then unchecking and still no effect. I've downgraded to v0.15.6 (93) for now.

EvanDorsky:
It looks like this issue was created by PR #1209.

I resolved the issue locally by wrapping this line: https://github.com/ianyh/Amethyst/blob/cbf9612b1342ccc60a9edbaaa1dc2ec92f51ecd1/Amethyst/Managers/WindowManager.swift#L662

with the appropriate preference check:
```Swift
if UserConfiguration.shared.followWindowsThrownBetweenSpaces() {
    window.focus()
}
```

This builds and runs locally and the preference is once again respected.

@dozzman (author of PR #1209) does this change break any desirable behavior? I will submit a PR soon.

ondvrsn:
I was hoping this fix would be included in 0.16.1 update. May I ask when this fix will be officially released?

PR #1333, opened 2022-10-28
Issue 1306: Fix "Follow thrown windows between spaces" setting not working

EvanDorsky:
This PR resolves Issue #1306 (Throw a window to other spaces will follow the window to that space even when the settings is unchecked.)

It looks to me like PR #1209 broke the "follow thrown windows between spaces" setting by adding a new call to `window.focus` that was not gated by a check for `UserConfiguration.shared.followWindowsThrownBetweenSpaces()`.

Adding this check restores the expected behavior. This code builds and runs locally.

Pinging @dozzman as the author of PR #1209, just want to make sure adding this check doesn't break anything (I can't imagine it would, though).

Also thanks @ianyh for making contributing easy – I was able to get the project up and running in no time.


ianyh (approved):
Nice catch.

PR #1358, opened 2023-01-08
feat toggle window fullscreen

Kane-Allan:
Adds a new keybind option, When pressed will set the current layout to fullscreen and save what it was before. When pressed again, will revert to previous layout.

I am not familiar with swift so apologies.

Just started using Amethyst and thought it was great but I really wanted this feature.

I have been testing it for a few days with no issues but probably needs testing in a few other scenarios, if it is even something you want to merge.


Amethyst/Managers/ScreenManager.swift
```diff
@@ -367,6 +368,26 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
...
+        }
+
+        if currentLayout == "fullscreen" {
+            previousLayout = previousLayout ?? layouts.first(where: {$0.layoutKey != "fullscreen"})?.layoutKey
+
+            if previousLayout == nil {
+                return
+            }
+
+            selectLayout(previousLayout!)
```
rickythefox: Why not use a `guard let` instead of force unwrapping?

Kane-Allan: Don't have a good reason for that other than I assumed using 

`guard let previousLayout = previousLayout ?? layouts.first(where: {$0.layoutKey != "fullscreen"})?.layoutKey else {
                return
            }` 

would create a local variable rather than asign to the previousLayout in the outer scope.

As I mentioned, not at all familiar with swift. Happy to change it

rickythefox: I would rewrite this as below to reduce the nesting and remove the force unwrapping. Swift is definitely not one of my primary languages either so maybe this can be done better.

```
    func toggleFullscreen() {
        guard let currentLayout = currentLayout?.layoutKey else {
            return
        }
        
        if (previousLayout != "fullscreen") {
            previousLayout = currentLayout
            selectLayout("fullscreen")
            return
        }
        
        guard let firstNonFullscreenLayout = layouts.first(where: {$0.layoutKey != "fullscreen"}),
              let previousLayout = previousLayout ?? firstNonFullscreenLayout else {
            return
        }

        selectLayout(previousLayout)
    }
```

Kane-Allan: Based on your psuedo code I have this, which may be better than my original implementation

```
func toggleFullscreen() {
        guard let currentLayout = currentLayout?.layoutKey else {
            return
        }

        if currentLayout != "fullscreen" {
            previousLayout = currentLayout
            selectLayout("fullscreen")
            return
        }

        guard let layout = previousLayout ?? layouts.first(where: {$0.layoutKey != "fullscreen"})?.layoutKey else {
            return
        }

        previousLayout = layout
        selectLayout(layout)
    }
```

ianyh (changes requested):
I think the idea behind this is reasonable, but I think it should be generalized to the select layout commands that already exist. We should keep track of the previous layout and toggle back if we are already on the layout when we use the select command.

Kane-Allan:
> I think the idea behind this is reasonable, but I think it should be generalized to the select layout commands that already exist. We should keep track of the previous layout and toggle back if we are already on the layout when we use the select command.

Are you suggesting this for any layout? like so?

```swift
func selectLayout(_ layoutString: String) {
        previousLayout = layoutString
        guard let currentLayout = currentLayout?.layoutKey else {
            return
        }
        
        let nextLayout = currentLayout == layoutString ? previousLayout : layoutString
        
        guard let layoutIndex = layouts.firstIndex(where: { $0.layoutKey == nextLayout }) else {
            return
        }

        setCurrentLayoutIndex(layoutIndex)
        setNeedsReflow(withWindowChange: .layoutChange)
    }
```

ianyh:
Yeah, that's the idea.

Amethyst/Managers/ScreenManager.swift
```diff
@@ -278,12 +279,21 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
     }
 
     func selectLayout(_ layoutString: String) {
-        guard let layoutIndex = layouts.firstIndex(where: { $0.layoutKey == layoutString }) else {
+        log.debug("layout is " + layoutString)
```
ianyh: Remove, please.

Amethyst/Managers/ScreenManager.swift
```diff
@@ -48,6 +48,7 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
     private var currentLayoutIndexBySpaceUUID: [String: Int] = [:]
     private var layoutsBySpaceUUID: [String: [Layout<Window>]] = [:]
     private var currentLayoutIndex: Int = 0
+    var previousLayout: String?
```
ianyh: `previousLayoutKey`

Amethyst/Managers/ScreenManager.swift
```diff
@@ -278,12 +279,21 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
     }
 
     func selectLayout(_ layoutString: String) {
-        guard let layoutIndex = layouts.firstIndex(where: { $0.layoutKey == layoutString }) else {
+        log.debug("layout is " + layoutString)
+
+        guard let currentLayout = currentLayout?.layoutKey else {
```
ianyh: `currentLayoutKey`

Amethyst/Managers/ScreenManager.swift
```diff
@@ -278,12 +279,21 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
...
+            return
+        }
+
+        let nextLayout = currentLayout == layoutString ? previousLayout : layoutString
```
ianyh: `nextLayoutKey`

Amethyst/Managers/ScreenManager.swift
```diff
@@ -278,12 +279,21 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
...
+
+        guard let layoutIndex = layouts.firstIndex(where: { $0.layoutKey == nextLayout }) else {
```
ianyh: I think there is a `first` function for getting the layout itself instead of getting the index.

ianyh (changes requested):

Amethyst/Managers/ScreenManager.swift
Kane-Allan: Can you explain some more what you mean here please? 

I did not change that part of the code and this function `setCurrentLayoutIndex(layoutIndex)` depends on the layoutIndex.

ianyh: Ah, you're right. Yeah, let's keep the existing code.

ianyh (approved):

DanielAsher:
Dope

PR #1366, opened 2023-01-15
Allow extending existing layouts

ianyh:


ianyh:
The translation of window frames to js wasn't being done correctly, but I'm worried about how many custom layouts were relying on the old format.

PR #1380, opened 2023-01-20
Switch to previous main window when main is selected

### LINKED ISSUE #1377: Make "Swap focused window with main window" usable when main window is focused
Currently the `mod1 + enter` shortcut doesn't do anything when the main window is already focused. I think it's a shame as it's a very convenient shortcut.

My proposal is to use the shortcut to swap the main window with the previous main window. This way you can easily toggle between two windows while working.

@ianyh would love to hear your thoughts on this.

### LINKED ISSUE #1378: Toggle focus between panes in Two Pane layout
In two pane layout, it is not possible to smoothly toggle focus between the main and secondary panes. Example, with more than 2 windows open: 
  1. Swap any open window to the main pane (⌘+⇧+Enter)
  2. Move focus cw/ccw (⌘+⇧+J/K)

After 2, the window initially occupying the secondary pane will be hidden behind a different window.

We need a solution to swap focus between main and secondary panes without affecting window order. I'd suggest something like what was proposed in #1234, e.g. pressing ⌘+⇧+M when the main pane is focused should toggle to the secondary pane.

It may be possible to simply reorder Amethyst's stack of windows following 1. above so that the window displayed on the secondary pane will receive focus after 2. This seems suboptimal as one of ⌘+⇧+J or ⌘+⇧+K would still surface a window from the bottom of the stack, but might be cleaner to implement.

rickythefox:
The idea is to save the previous main windows on a stack and switch to the previous main window when invoking the "Swap focused window with main window" and the main window is already focused. If there are no previous main windows then the first clockwise window is used instead.

I see this as a draft PR and would like to hear what you think about the approach. No support for multiple desktops / screens yet.

Closes #1377, closes #1378

### DIFF
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -688,6 +688,17 @@ extension WindowManager: WindowTransitionTarget {
 
         return (screenManagerIndex == 0 ? screens.screenManagers.count - 1 : screenManagerIndex - 1)
     }
+
+    func lastMainWindowForCurrentSpace() -> Window? {
+        guard let screenManager = screens.focusedScreenManager(),
+              let currentFocusedSpace = CGSpacesInfo<Window>.currentFocusedSpace(),
+              let lastMainWindow = windows.lastMainWindows[currentFocusedSpace.id]
+        else {
+            return nil
+        }
+        log.debug("Get last main window on space \(currentFocusedSpace.id) = \(lastMainWindow?.title() ?? "nil")")
+        return lastMainWindow
+    }
 }
 
 // MARK: Focus Transition
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -696,7 +696,6 @@ extension WindowManager: WindowTransitionTarget {
         else {
             return nil
         }
-        log.debug("Get last main window on space \(currentFocusedSpace.id) = \(lastMainWindow?.title() ?? "nil")")
         return lastMainWindow
     }
 }
--- a/Amethyst/Managers/WindowTransitionCoordinator.swift
+++ b/Amethyst/Managers/WindowTransitionCoordinator.swift
@@ -31,6 +31,7 @@ protocol WindowTransitionTarget: class {
     func activeWindows(on screen: Screen) -> [Window]
     func nextScreenIndexClockwise(from screen: Screen) -> Int
     func nextScreenIndexCounterClockwise(from screen: Screen) -> Int
+    func lastMainWindowForCurrentSpace() -> Window?
 }
 
 class WindowTransitionCoordinator<Target: WindowTransitionTarget> {
@@ -54,14 +55,18 @@ class WindowTransitionCoordinator<Target: WindowTransitionTarget> {
             return
         }
 
-        if focusedIndex == 0 && target?.currentLayout()?.layoutKey == TwoPaneLayout<Window>.layoutKey {
-            // If main window is focused and layout is two-pane - swap the two top-most windows, keep focus on the main window
-            target?.executeTransition(.switchWindows(focusedWindow, windows[1]))
-            windows[1].focus()
+        if focusedIndex == 0 {
+            guard let lastMainWindow = target?.lastMainWindowForCurrentSpace() else {
+                return
+            }
+            log.debug("Swapping to last main window \(lastMainWindow.title())")
+            target?.executeTransition(.switchWindows(focusedWindow, lastMainWindow))
+            lastMainWindow.focus()
             return
         }
 
         if focusedIndex != 0 {
+            log.debug("Swapping with main window \(windows[0].title())")
             // Swap focused window with main window if other window is focused
             target?.executeTransition(.switchWindows(focusedWindow, windows[0]))
         }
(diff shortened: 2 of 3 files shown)


ianyh:
Two thoughts:
1. We already have a concept of last focused window as a part of the focus coordinator. We could probably create an analogous concept of a last swapped window.
2. I would make this behavior a part of `swapFocusedWindowToMain()`, I think. If the main window is already focused then grab the last swapped window.

rickythefox:
@ianyh Updated to work more like the last focused window. We need to keep track of which space the main window belongs to as you probably want to switch to last main window in the same space - otherwise you get pretty strange jumps between spaces.

ianyh:
Approach seems sound. Nice catch on the consideration of spaces.

rickythefox:
Cleaned up a bit and removed the log statements. I'm not a fan of the way this behaves in 2-pane layout, but I honestly don't see how we can make it better:
1. The way 2-pane layout behaves is consistent with XMonad (checked)
2. In XMonad invoking the "swap window to main" shortcut when main is focused does nothing. With this PR we do get the previous main window back. See https://drive.google.com/file/d/1O_ofYbgg8iuw4uRbxhqnrkmcBgqwOrBO/view?usp=drivesdk .

If this behaviour is acceptable I'll remove the draft label. If not - I'm open to suggestions re. how this should work in 2-pane. @brush701 maybe you want to provide some input here?

brush701:
It's a bit hard to tell exactly what keys are being pressed in the video, but based on the code this seems pretty good to me. Curious what you're not a fan of?

rickythefox:
@brush701 my bad, here's a video with keystrokes. I think the last switch looks a bit funny, but I'm not used to two-pane layout. Does this look ok to you?

https://drive.google.com/file/d/1gRngVSt1ULe9tmWiJz3PGDfvgh_N9GX0/view?usp=drivesdk

brush701:
Thanks for the updated video. Two-pane does seem a bit wonky to me. It seems like in two pane with the main pane focused, hitting Mod+Enter swaps the two windows between panes but the second pane never takes focus? That is the default behavior I see on my system with the current release build, but I might be missing something. 
 
It would be most helpful to have a mechanism to keep the windows in place but shift the focus to the secondary pane. For example, the current behavior is: 
	1.	I have 4 windows open, 2 pane layout. Window 1 is in the main pane, window 2 is in the second.
	2.	I hit Mod1+K, which focuses window 2.
	3.	I hit Mod1+K again, bringing window 3 into focus in the second pane
	4.	I hit Mod1+M, focusing the main window. Window 3 is still in the second pane
	5.	I hit Mod1+K. Instead of focusing on window 3, it focuses on window 2. 
 
I think the fundamental problem here is that in 2 pane, changing focus for windows in the secondary pane is actually the same as swapping b/c it produces a different visual layout of visible windows on screen. That might be the cleanest way to implement a change here? Perhaps something like this at the end of `FocusTansitionCoordinator.moveFocus[Counter]Clockwise()`
```
 if focusedWindow.screen().currentLayout()?.layoutKey == TwoPaneLayout<Window>.layoutKey {
            if windows[0] != focusedWindow {
                windows.swap(window: focusedWindow, withWindow: windowToFocus)
            }
        }
```
There might be some issues not calling `markAllScreensForReflow()`, but that's inaccessible from this class. Might be an even better way to plumb all of this through the `WindowManager`, which feels like the right level of abstraction, but it wasn't immediately clear to me how one would do that. I'd test it myself but I'm not really a swift dev and don't have my tooling set up all the way to build it.

rickythefox:
It seems like the same wonkiness is present in xmonad: https://drive.google.com/file/d/1b9oBmiSHf0GHkM6UOgKM2bE8tvaE9Y-R/view?usp=drivesdk .

I'm not sure we would want to change the behaviour here as it would depart from the screen manager we're tying to replicate. @ianyh , what's your take on this?

@brush701 I'm by no means a Swift developer myself but the tooling is pretty easy to set up and use. The only issue is to avoid window chaos when testing the builds. That's why I use a VM which works reasonably well but then I have to debug using log statements which brings me back to the 90s. :)

ianyh:
Is the issue the focus after the swap? That is to say, is the desired behavior that when swapping _back_ after the original main pane swap the now-main pane window be focused?

brush701:
For me the issue is that there's currently no combination of keys I can press to reliably toggle focus back and forth between main and secondary panes because of how `focusMain` interacts with `focusCW/CCW`. Using mod1+j/k requires multiple keystrokes to move from secondary to main pane if you have more than a couple of windows open. If you throw in a mod1+m, then it "loses track" of what window is supposed to be on the secondary pane. Desired behavior would be that after mod+m, mod+j would shift focus to the window currently displayed in the second pane. Instead, it currently shifts focus to either `windowManager.windows[1]` or `[-1]`, which is not the same thing.

rickythefox:
I think the is an issue that needs to be adressed in another PR. The behaviour right now seems consistent with xmonad (see my videos above). 

I'm removing the draft tag from this PR and will see if I can submit another with a proposed behaviour change for the 2-pane layout. Alternatively maybe a custom layout can be created using the js integration.

brush701:
Sounds good. I think I've got my desired behavior running locally, so I'll open a separate PR for that.

ianyh (approved):
I think I'm fine with xmonad parity, and we can consider the other PR an improvement.

PR #1388, opened 2023-01-23
Allow handling apps just by windows title without bundle identifier (Android Emulator Example)

jmalczak:
Currently android emulator doesn't have bundle identifier and it's not available as app package so there is no way to add it to floating list. This pull request allow providing * as bundle identifier and matching just by window title.

Screenshot 2023-01-23 at 21 20 07
Screenshot 2023-01-23 at 21 22 57

### DIFF
--- a/Amethyst/Preferences/UserConfiguration.swift
+++ b/Amethyst/Preferences/UserConfiguration.swift
@@ -533,7 +533,7 @@ class UserConfiguration: NSObject {
     func runningApplicationFloatingBundle(_ runningApplication: BundleIdentifiable) -> FloatingBundle? {
         let floatingBundles = self.floatingBundles()
 
-        let bundleIdentifier = runningApplication.bundleIdentifier ?? "*"
+        let bundleIdentifier = runningApplication.bundleIdentifier ?? ""
 
         for floatingBundle in floatingBundles {
             if floatingBundle.id.contains("*") {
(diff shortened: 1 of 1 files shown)


Amethyst/Preferences/UserConfiguration.swift
```diff
@@ -533,13 +533,11 @@ class UserConfiguration: NSObject {
     func runningApplicationFloatingBundle(_ runningApplication: BundleIdentifiable) -> FloatingBundle? {
         let floatingBundles = self.floatingBundles()
 
+        let bundleIdentifier = runningApplication.bundleIdentifier ?? "*"
```
ianyh: Maybe this should just be the empty string? `.*` should still match that.

ianyh (changes requested):

Amethyst/Preferences/UserConfiguration.swift
jmalczak: To be honest I am so new to this project that I don't really have opinion about that :) If you prefer it to be empty string, I can test it if it works and change accordingly?

ianyh: I think conceptually I would prefer an empty string.

jmalczak: Done.

ianyh (approved):

Arsh-78:
This Works man Thanks a lot

PR #1390, opened 2023-01-27
Main window toggle

brush701:
This PR proposes changing the behavior of the `Move focus to main window` command, default  ⌥⇧M, if the main window is already focused. The current default behavior is to do nothing. This PR changes the behavior to switch focus back to the last focused window. This allows the user to easily toggle between the main window and another, which is especially useful in the Two Pane Layout.


ianyh:
Generally looks good, but could you rebase out the 0.17.0b commit?

brush701:
Should be all set

ianyh (approved):

PR #1395, opened 2023-01-30
Fix crash when full-screening a window

rickythefox:
When full-screening a window `activeWindows ` returns an empty array which leads to a crash.


ianyh (approved):
Nice catch.

PR #1397, opened 2023-01-31
When adding windows use the previous main as last focused

rickythefox:
When adding a new window with the "Send new windows to main pane" setting active, we should save the previous main window.

Sorry for spamming PRs on this topic.

### DIFF
--- a/Amethyst/Managers/Windows.swift
+++ b/Amethyst/Managers/Windows.swift
@@ -58,11 +58,25 @@ extension WindowManager {
             return screenWindows
         }
 
+        func activeWindowOnCurrentScreen(atIndex: Int) -> Window? {
+            guard let focusedWindow = Window.currentlyFocused(),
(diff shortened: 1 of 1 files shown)


Amethyst/Managers/Windows.swift
```diff
@@ -58,18 +58,41 @@ extension WindowManager {
             return screenWindows
         }
 
+        func activeWindowsOnCurrentScreen() -> (Space, [Window])? {
```
ianyh: I understand the convenience of returning the tuple, but I would prefer the space call to be separate.

ianyh (changes requested):

Amethyst/Managers/Windows.swift
rickythefox: Done

ianyh (approved):

PR #1433, opened 2023-02-22
Add support for throwing to spaces 11 and 12

timvisher:
I run with 12 spaces. The only trick is to map throwing to the Function keys rather than to the numeric keys.

I'm not sure whether this works because I can't seem to run the tests locally. ~I was trying to follow [.travis.yml](https://github.com/timvisher/Amethyst/blob/51d2478008d21e04c64ccb5831da14aa35854fd0/.travis.yml#L8) but running bundle install complains that I don't have Ruby 2.4.1 installed and when I try to `rbenv install 2.4.1` it tells me its EOL.~ Welp that's what happens when you check out `master` rather than `development`. xD

So I'll throw this up as a Draft and see what happens.

### DIFF
--- a/Amethyst/default.amethyst
+++ b/Amethyst/default.amethyst
@@ -166,6 +166,8 @@
         "mod": "mod2",
         "key": "0"
     },
+    "throw-space-11": [],
+    "throw-space-12": [],
     "throw-space-left": {
         "mod": "mod2",
         "key": "left"
--- a/Amethyst/default.amethyst
+++ b/Amethyst/default.amethyst
@@ -166,8 +166,6 @@
         "mod": "mod2",
         "key": "0"
     },
-    "throw-space-11": [],
-    "throw-space-12": [],
     "throw-space-left": {
         "mod": "mod2",
         "key": "left"
--- a/Amethyst/Events/HotKeyManager.swift
+++ b/Amethyst/Events/HotKeyManager.swift
@@ -385,7 +385,7 @@ class HotKeyManager<Application: ApplicationType>: NSObject {
         hotKeyNameToDefaultsKey.append(["Throw focused window to space left", CommandKey.throwSpaceLeft.rawValue])
         hotKeyNameToDefaultsKey.append(["Throw focused window to space right", CommandKey.throwSpaceRight.rawValue])
 
-        (1...10).forEach { spaceNumber in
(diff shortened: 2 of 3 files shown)


ianyh:
You need to rebase off of the most current branch.

timvisher:
Hmm… I'm still not able to install Ruby 3.0.5 based on the actual latest commit.

```
$ rbenv install -l
2.6.10
2.7.6
3.0.4
3.1.2
jruby-9.3.8.0
mruby-3.1.0
picoruby-3.0.0
rbx-5.0
truffleruby-22.2.0
truffleruby+graalvm-22.2.0

Only latest stable releases for each Ruby implementation are shown.
Use 'rbenv install --list-all / -L' to show all local versions.
```

Admittedly I'm neither a ruby dev nor a swift dev so… LOL. Maybe I'm not the right person to make this change.

ianyh:
You might need to update ruby-build if you did that through homebrew.

timvisher:
Seems like that's working reasonably well. Should I need to change `pod install` → `bundle exec pod install` locally?

timvisher:
Seems promising?

```
$ xcodebuild -workspace Amethyst.xcworkspace -scheme Amethyst clean test
…
** TEST SUCCEEDED **
```

timvisher:
Ah. Looks like there was a `bin/setup.sh` that would've done a lot of this for me. xD

timvisher:
I'm not sure if there's a way to create a local development version of the application that I can actually run to test but I think this is ready for review.

Amethyst/default.amethyst
```diff
@@ -166,6 +166,8 @@
         "mod": "mod2",
         "key": "0"
     },
+    "throw-space-11": [],
```
ianyh: You should actually just remove these. They will still default to not being set.

ianyh (changes requested):

ianyh:
Also,
> Admittedly I'm neither a ruby dev nor a swift dev so… LOL. Maybe I'm not the right person to make this change.

You did fine. 🙂

Amethyst/default.amethyst
timvisher: kk back to you. :)

ianyh (approved):

PR #1529, opened 2023-07-31
feat: add delete keyboard key to non-printable characters map

tpucci:
This PR adds "delete" to the recognizable keyboard non-printable characters.

We can now use "delete" in configuration files.


tpucci:
And thanks you for this wonderful piece of software, I can not work without amethyst ❤️

ianyh (approved):
Cool! Thanks.

PR #1551, opened 2023-09-17
Proxies PanedLayout function calls to extended layouts

### LINKED ISSUE #1550: Shrink/Extend main pane does not work for extended custom layouts
**Describe the bug**
If a custom layout extends a layout that uses a main pane  (e.g. 3Column Middle), then Shrink/Extend main pane & Increase/Decrease main pane count shortcuts do not have any affect.

**To Reproduce**
The following custom layout can be used to reproduce the issue.

```js
function layout() {
  return {
    name: "test",
    extends: "middle-wide",
    getFrameAssignments: (windows, screenFrame, state, extendedFrames) => {
      return extendedFrames.reduce((frames, extendedFrame) => {
        return { ...frames, [extendedFrame.id]: extendedFrame.frame };
      }, {});
    }
  };
}
```

1. Load the above custom layout
2. Try to shrink the main pane

**Expected behavior**
The above custom layout extends the `3Column Middle` layout, and the Shrink/Extend main pane & Increase/Decrease main pane count shortcuts should work normally.

**Versions:**
 - macOS: Ventura 13.5
 - Amethyst: 0.20.0 (108)
 
**Debug Info**
```
ersion: 0.20.0 (108)

OS version: Version 13.5 (Build 22G74)

Screens:
	(0.0, 0.0, 5120.0, 1440.0) [(0.0, 0.0, 5120.0, 1440.0)]
	(1910.0, -1080.0, 1920.0, 1080.0) [(1910.0, 1440.0, 1920.0, 1080.0)]

Configuration:
smart-window-margins: 0
new-windows-to-main: 0
mod2: (
    option,
    shift,
    control,
    command
)
float-small-windows: 1
screen-padding-left: 0
mod1: (
    option,
    shift,
    command
)
window-minimum-width: 0
restore-layouts-on-launch: 1
use-canary-build: 0
window-margins: 1
mouse-resizes-windows: 1
floating-is-blacklist: 1
debug-layout-info: 0
enables-layout-hud: 1
follow-space-thrown-windows: 1
ignore-menu-bar: 0
floating: (
)
window-resize-step: 5
window-max-count: 0
enables-layout-hud-on-space-change: 0
layouts: (
    test,
    "middle-wide"
)
screen-padding-bottom: 0
window-minimum-height: 0
screen-padding-right: 0
mouse-swaps-windows: 1
focus-follows-mouse: 0
mouse-follows-focus: 0
screen-padding-top: 0
window-margin-size: 10
```

Henkru:
Closes #1550.

This PR extends the Custom Layout class to proxy PanedLayout functions to the extended layout.

### DIFF
--- a/Amethyst/Layout/Layouts/CustomLayout.swift
+++ b/Amethyst/Layout/Layouts/CustomLayout.swift
@@ -27,7 +27,7 @@ private enum LayoutExtension<Window: WindowType> {
     case layout(Layout<Window>)
 }
 
-class CustomLayout<Window: WindowType>: StatefulLayout<Window> {
+class CustomLayout<Window: WindowType>: StatefulLayout<Window>, PanedLayout {
     typealias WindowID = Window.WindowID
 
     private enum CodingKeys: String, CodingKey {
@@ -48,6 +48,9 @@ class CustomLayout<Window: WindowType>: StatefulLayout<Window> {
 
     private let key: String
     private let fileURL: URL
+
+    private(set) var mainPaneCount: Int = 1
+    private(set) var mainPaneRatio: CGFloat = 1.0
 
     private lazy var context: JSContext? = {
         guard let context = JSContext() else {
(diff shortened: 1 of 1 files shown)


ianyh (changes requested):
This is cool, but the one point of feedback is that I wonder if the pane layout commands should act as defaults for the custom commands that can be overridden by explicit definitions.

My concern is that if you have multiple custom layouts you may have muscle memory for the custom commands, and I think it would be jarring to have to switch to the other set of commands to get this to work.

Henkru:
@ianyh Thanks for the feedback. Do I interpret your message correctly as meaning: custom command 1 = shrink, custom command 2 = expand, etc.?

I understand the point you're making about it being jarring, but I'd like to elaborate on it further. Firstly, the capability to extend built-in or existing layouts is excellent. Amethyst offers a comprehensive set of built-in layouts, so my assumption is that tweaking existing layouts is more common than creating new ones from scratch. However, this conclusion is based on my own perception, so I could be mistaken.

In the scenario mentioned above, it seems non-intuitive to, for instance, shrink the main pane using the shortcut designated for a custom command (if the selected custom layout extends a built-in one).

For example, in my case, I use a 49" ultra-wide monitor and prefer the 3-column layout. However, when I have only one window open, I want it in the middle column without stretching it to fullscreen. This is easily achievable by extending the built-in 3-column layout and modifying the behavior for that specific case. I might use the two-pane layout if my laptop is not connected to an external monitor. Here's where the non-intuitive aspect comes in: I would then have different key shortcuts for shrinking the pane, depending on the layout in use (the built-in two-pane layout: "shrink main pane," or my custom 3-column layout: "custom command 1"). In my view, I shouldn't need to be aware of the current layout to use the correct shortcut to shrink the pane.

Any thoughts on that?

In addition, since you know better the overall architecture, could there be any corner cases where PanedLayout is handled differently compared to "normal" layouts? Since my modifications changed `CustomLayout` to implement `PanedLayout` as well.

ianyh:
I've seen a lot of layouts that are made from scratch, but that may be a consequence of not having had extensions at the start. I wonder if the solution here isn't to proxy the commands, but rather to proxy the pane state since that is fairly well defined by the protocol? Or maybe have the option for extensions of pane layouts to proxy to the native commands from a custom command? e.g.,
```js
{
    ...
    command1: "expandMainPane"
    ...
}
```

garo:
As I'm writting a custom layout, which is not extending any layout, I would love to be able to adjust my layout using the native `Shrink main pane` and `Extend main pane` shortcuts, so that I can use the same keyboard shortcuts regardless which layouts I'm using.

HelloThisIsFlo:
Oh yes @garo , that would be great indeed!

I have almost the exact same use-case. I'm using 2 custom commands to replicate `increase main pane count` and `decrease main pane count`. I'd love to be able to use the same shortcut as the default layouts. My solution for now is not ideal: I simply ... never use the default layouts and rely exclusively on custom layouts (specifically because of this shortcut mismatch)

ncasdpo:
@Henkru Are you able to accept ianyh's change so this can be merged?

ianyh:
Superseded by https://github.com/ianyh/Amethyst/pull/1735

PR #1561, opened 2023-10-01
add profiles management

bartlomiejwojsa:
This feature helps to manage shortcuts profiles. Especially helpful when frequently switching from one keyboard type to another.


ianyh:
@bartlomiejwojsa would you be willing to update this PR?

ianyh:
Having thought about this quite a bit, I think I am going to close this. I like the idea of it, but it increases the surface area in a way that will be hard to maintain. Does a new feature need to have base functionality, custom layout, dotfile, profile, etc.?

Issue #1567, opened 2023-10-24
Window thrown to a space loses focus


dzanotto:
**Describe the bug**
When a window is thrown to a space with "Follow thrown windows between spaces" enabled, the window loses focus.

**Applications:**
Every application

**To Reproduce**
1. Set up at least two spaces
2. On the first space open at least one window and focus it
3. On the second space open a window and focus it
4. Throw the window to the first space using the keyboard shortcut
5. The focus switches from the window being moved to the one previously focused on the destination space

**Expected behavior**
The window thrown to the destination space should keep the focus

**Versions:**
 - macOS: Sonoma 14.0
 - Amethyst: 0.20.0 (108)

cyberthirst:
I've been experiencing too (even before Sonoma). Currently on the same versions as OP and the bug persists.

LouisKraemer:
Same here, can reproduce each time. Can provide additional informations if needed.

cyberthirst:
Probably related to this recent commit? https://github.com/ianyh/Amethyst/commit/4d0e820beb25f4bdb89088326a470a9132d89ccb

I can see that the **focus** function was modified. Seems very messy haha :D Don't really know how it should behave without further debugging.

dzanotto:
I think it's not related to that commit.
Plus, I can reproduce the same behaviour on Sonoma even with older versions (I've tried backwards until 0.15.6, and I'm sure that that version was working before Mac OS update)

MichalMoravik:
Same issue here, any update on this please?

ianyh:
This is weird. I did a bunch of testing, and it seems like the function just doesn't behave correctly anymore. It still loses focus for every case I tried; disabling following to the space, disabling the new raise code, disabling raising at all, etc. Nothing seems to work. It seems like yabai sidesteps it with the Dock injection, but I'm not sure how to do it here.

ianyh:
Could someone(s) look at the video in the PR? I could use feedback on how janky is too janky. https://github.com/ianyh/Amethyst/pull/1586

dzanotto:
> This is weird. I did a bunch of testing, and it seems like the function just doesn't behave correctly anymore. It still loses focus for every case I tried; disabling following to the space, disabling the new raise code, disabling raising at all, etc. Nothing seems to work. It seems like yabai sidesteps it with the Dock injection, but I'm not sure how to do it here.

Additional information: yabai works even without Dock injection (so without disabling System Integrity Protection).
[...]

PR #1586, opened 2024-01-04
Only perform raise ax action when focusing

ianyh:
Potential solution for #1567. It looks super janky—when the space transition finishes the window flickers in and out—but is... functional? I'm kind of torn.


ianyh:
https://github.com/ianyh/Amethyst/assets/212270/3adcf57a-04a6-41b9-ad99-3a14de582643

Example of what the flickering looks like.

ianyh:
Okay, I found the actual problem. Silica's focus code was also doing an application activate in addition to the raise action.

PR #1604, opened 2024-02-18
Add Floating Grid Layout

MutatingFunc:
For more information about the thought behind this approach:
https://mutatingfunc.github.io/blog/2023-11-03-macos-amethyst-window-snapping/

Short version - iPad-style grid alignment.

Longer version - Floating layout, but windows are snapped to a grid to make it easier to align them.

[Demo build download](https://github.com/MutatingFunc/Amethyst/releases/tag/floating-grid-demo-2) (updated 2025.03.21)


ianyh:
Sorry I haven't had a chance to look at this! It seems neat, I just need to think about it a bit more.

MutatingFunc:
I've added a demo build here:
~https://github.com/MutatingFunc/Amethyst/releases/tag/floating-grid-demo~

MutatingFunc:
Made a few changes following the rebase, as the approach I was using to detect window position & size changes stopped working. This build integrates directly with the window resizing and repositioning handlers to trigger the grid snapping, so it should be much more robust.

https://github.com/MutatingFunc/Amethyst/releases/tag/floating-grid-demo-2

Amethyst.xcodeproj/project.pbxproj
```diff
@@ -1026,7 +1011,7 @@
 				CURRENT_PROJECT_VERSION = 121;
 				DEAD_CODE_STRIPPING = YES;
 				DEFINES_MODULE = YES;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = VKFDYMU9HJ;
```
ianyh, line 1014: Can you revert these?

Amethyst/Managers/WindowManager.swift
```diff
@@ -622,19 +627,25 @@ extension WindowManager: ApplicationObservationDelegate {
...
-
-        guard
-            let screenManager: ScreenManager<WindowManager<Application>> = focusedScreenManager(),
-            let layout = screenManager.currentLayout,
-            layout is PanedLayout
-        else {
+        
+        guard layout is PanedLayout else {
+            if layout is FloatingGridLayout, case .doneDragging = mouseStateKeeper.state {
+                screenManager.setNeedsReflow(withWindowChange: .layoutChange)
```
ianyh: I don't think `.layoutChange` is right. That is for when the currently selected layout changes. We may need a different one for specifically window resizes.

ianyh (changes requested):
Overall seems like a cool change. I think some of this could be extended to support custom layouts, but I am okay with this being a native layout. It seems pretty easy to maintain.

MutatingFunc:
Hey, thanks for the feedback, makes sense! I think this came through day 1 of my 4 week holiday abroad though so I’ll get back to this when I’m at my laptop properly again in May!

MutatingFunc:
Hey, I'm back! I've removed my team ID from the Git history for security/privacy, and generalised the window resize event. However, the resize event is still only triggered for the Floating Grid layout, for efficiency. Potentially it could be worth looking at generalising this down the road to a protocol if it makes sense for other layout types to observe the event.

MutatingFunc:
Does this now look good to merge?

rwjack:
This looks very cool!

PR #1618, opened 2024-03-21
Match identifiers first for floating windows

tuyafeng:
Hi ianyh,

Thank you for your hard work! Your app is truly fantastic. However, I've encountered a minor issue. While using it, I noticed that the behavior of floating windows seems to be influenced by the order of rules. For instance, consider the following rules:

```
* (With title rules, such as "MI 6")
com.apple.finder
```

In this setup, the first rule takes precedence, causing Finder windows to be unable to float properly. However, if I rearrange the rules like so:

```
com.apple.finder
* (With title rules, such as "MI 6")
```

Everything works perfectly fine. Therefore, I made a small adjustment accordingly. Since I'm not proficient in Swift, please feel free to correct any mistakes. Once again, thank you for creating such a wonderful app!

### DIFF
--- a/Amethyst/Preferences/UserConfiguration.swift
+++ b/Amethyst/Preferences/UserConfiguration.swift
@@ -555,7 +555,7 @@ class UserConfiguration: NSObject {
     }
 
     func runningApplicationFloatingBundle(_ runningApplication: BundleIdentifiable) -> FloatingBundle? {
-        let floatingBundles = self.floatingBundles()
+        let floatingBundles = self.floatingBundles().sorted { Int($0.id.contains("*")) < Int($1.id.contains("*")) }
 
         let bundleIdentifier = runningApplication.bundleIdentifier ?? ""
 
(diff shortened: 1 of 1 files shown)


Amethyst/Preferences/UserConfiguration.swift
```diff
@@ -555,7 +555,7 @@ class UserConfiguration: NSObject {
     }
 
     func runningApplicationFloatingBundle(_ runningApplication: BundleIdentifiable) -> FloatingBundle? {
-        let floatingBundles = self.floatingBundles()
+        let floatingBundles = self.floatingBundles().sorted { Int($0.id.contains("*")) < Int($1.id.contains("*")) }
```
ianyh: This is a good find, but for now could we just move the universal wild card `"*"` to the end of the list?

ianyh (changes requested):

tuyafeng:
I adjusted the judgment to move the wild card to the end.

PR #1762, opened 2025-04-01
Fixing the middle-wide layout focus sequence

### LINKED ISSUE #1189: Clockwise mode is inverted for 3-Column layouts
**Describe the bug**
In many layouts clockwise and counter clockwise are trivial to grasp, but what about the 3 column layout? I would expect that if I move focus or swap window clockwise, I would go to the right, the rotational axis should be on the bottom of the screen, because that feels natural. Unfortunately, this is not the case (at least for 3 Column Middle layout). When I move focus clockwise the window to the left will be focused, which feels unnatural to me.

**Applications:**
Any

**To Reproduce**

- Select the 3 Column Middle layout
- Move focus clockwise when the main pane is in focus

**Expected behavior**
Right window should get focus.

**Versions:**
 - macOS: 12.1 (21C52)
 - Amethyst: 0.15.5 (92)
 
**Debug Info**
```
$ /Applications/Amethyst.app/Contents/MacOS/Amethyst --debug-info
Version: 0.15.5 (92)

OS version: Version 12.1 (Build 21C52)

Screens:
	(0.0, 0.0, 2048.0, 1280.0) [(0.0, 0.0, 2048.0, 1280.0)]

Configuration:
window-margin-size: 5
window-resize-step: 5
use-canary-build: 0
floating-is-blacklist: 1
float-small-windows: 1
window-minimum-width: 0
window-margins: 1
enables-layout-hud: 1
mod2: (
    option,
    shift,
    control
)
floating: (
        {
        id = "com.apple.finder";
        "window-titles" =         (
        );
    }
)
ignore-menu-bar: 0
debug-layout-info: 0
mod1: (
    option,
    shift
)
screen-padding-bottom: 0
new-windows-to-main: 0
enables-layout-hud-on-space-change: 0
layouts: (
    "tall-right",
    wide,
    column,
    "two-pane",
    "middle-wide",
    bsp
)
screen-padding-right: 0
mouse-follows-focus: 0
mouse-swaps-windows: 1
mouse-resizes-windows: 1
smart-window-margins: 0
screen-padding-top: 0
screen-padding-left: 0
focus-follows-mouse: 0
follow-space-thrown-windows: 0
window-minimum-height: 0
restore-layouts-on-launch: 1
```

**Additional context**
The reason why it feels odd is because I use control+option+J for counter-clockwise and control+option+L for clockwise. If my bug is rather a subjective opinion I would be happy if the rotational axis position can be configured for layouts that have either only rows or only columns. On layouts that feature all orientations the rotational axis is in the middle obviously.

kelvinst:
Fixes #1189 

Basically, it goes in the wrong direction because the left column is marked as the "secondary", and when going clockwise it goes in the sequence `main` -> `secondary` -> `terciary`... So if we just switch secondary to `right` and terciary to `left`, it should go in the right direction.

### DIFF
--- a/Amethyst/Layout/Layouts/ThreeColumnLayout.swift
+++ b/Amethyst/Layout/Layouts/ThreeColumnLayout.swift
@@ -54,7 +54,7 @@ struct TriplePaneArrangement {
         self.panePosition = {
             switch mainPane {
             case .left:   return [.main: .left, .secondary: .middle, .tertiary: .right]
-            case .middle: return [.main: .middle, .secondary: .left, .tertiary: .right]
+            case .middle: return [.main: .middle, .secondary: .right, .tertiary: .left]
             case .right:  return [.main: .right, .secondary: .left, .tertiary: .middle]
             }
         }()


ianyh:
The problem with changing it now is that it would change the behavior out from under people already using it. I hate to say it, but maybe just add an option to that enum for a mirrored middle or something like that.

kelvinst:
I see. Yeah, another option looks like a good option, should I change this PR to do that then?

ianyh:
> I see. Yeah, another option looks like a good option, should I change this PR to do that then?

Yeah, if you make that change I can take a look.

rwjack:
In that case, could we just have another Layout named `3Column Middle (Fill Right first)` - or something similar, not to break the layout for users that got used to the current workings of it.

PR #1774, opened 2025-04-22
Allow accessory apps again

ianyh:


ianyh:
[Amethyst.zip](https://github.com/user-attachments/files/19841714/Amethyst.zip)

PR #1789, opened 2025-06-16
fix: Mouse cursor doesn't follow correctly when swapping windows

klesh:
When swapping windows, the mouse cursor would stay in the previous position of the window.
This PR fixes the problem by delaying the focus call


klesh:
Hi, please take a look at the fix when you find time.
I want to submit other fixes.

ianyh (approved):

PR #1804, opened 2025-08-20
Add Layouts Submenu to Application Menu

### LINKED ISSUE #1080: Make Layouts available from menus
It is too difficult to change layouts.

Would like to be able to switch the layout via the menus under the app icon.

I have shortcuts setup to shuffle through the configured layouts, but I want to jump directly to the one I want.

**Additional context**
That is all :)

rgeraskin:
### Description
This PR introduces a new layouts submenu in the application menu, providing users with quick access to layout options directly from the menu bar.

Closes: https://github.com/ianyh/Amethyst/issues/1080

### Changes Made
- **AppDelegate.swift**: Added 87 lines of code to implement the layouts submenu functionality
- **MainMenu.xib**: Updated the interface builder file to include the new submenu structure

### What's New
- ✨ New layouts submenu in the application menu
- 🎯 Enhanced user experience with easier layout access
- 📱 Improved menu navigation for layout management

### Files Modified
- `Amethyst/AppDelegate.swift` (+87 lines)
- `Amethyst/Base.lproj/MainMenu.xib` (+7 lines)

### DIFF
--- a/Amethyst/AppDelegate.swift
+++ b/Amethyst/AppDelegate.swift
@@ -28,6 +28,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     @IBOutlet var versionMenuItem: NSMenuItem?
     @IBOutlet var startAtLoginMenuItem: NSMenuItem?
     @IBOutlet var toggleGlobalTilingMenuItem: NSMenuItem?
+    @IBOutlet var layoutsMenuItem: NSMenuItem?
 
     private var isFirstLaunch = true
 
@@ -71,6 +72,11 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         hotKeyManager = HotKeyManager(userConfiguration: UserConfiguration.shared)
 
         hotKeyManager?.setUpWithWindowManager(windowManager!, configuration: UserConfiguration.shared, appDelegate: self)
+
+        // Populate layouts menu now that windowManager is initialized
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
+            self.populateLayoutsMenu()
+        }
     }
 
     override func awakeFromNib() {
@@ -93,6 +99,11 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         toggleGlobalTilingMenuItem?.title = "Disable"
 
         startAtLoginMenuItem?.state = (LoginServiceKit.isExistLoginItems(at: Bundle.main.bundlePath) ? .on : .off)
+
+        // Set up layouts menu delegate to refresh when opened
+        layoutsMenuItem?.submenu?.delegate = self
+
+        populateLayoutsMenu()
     }
 
     func applicationDidBecomeActive(_ notification: Notification) {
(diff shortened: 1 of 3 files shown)


Amethyst/AppDelegate.swift
```diff
@@ -173,6 +184,66 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+        let availableLayouts = LayoutType<SIApplication.Window>.availableLayoutStrings()
+
+        // Get current layout
+        let focusedScreenManager = windowManager?.focusedScreenManager()
+        var currentLayoutKey = focusedScreenManager?.currentLayout?.layoutKey
+
+        // If no focused screen manager, fallback to the first screen manager
+        if focusedScreenManager == nil, let firstScreenManager = windowManager?.screenManager(at: 0) {
+            currentLayoutKey = firstScreenManager.currentLayout?.layoutKey
+        }
```
rgeraskin: not sure that this is the correct way to find ScreenManager

rgeraskin:
rebased on development branch

Amethyst/AppDelegate.swift
```diff
@@ -93,6 +99,11 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         toggleGlobalTilingMenuItem?.title = "Disable"
 
         startAtLoginMenuItem?.state = (LoginServiceKit.isExistLoginItems(at: Bundle.main.bundlePath) ? .on : .off)
+
+        // Set up layouts menu delegate to refresh when opened
+        layoutsMenuItem?.submenu?.delegate = self
```
ianyh: It might be better to do this when the main menu is opened, so that the list is available before the submenu is accessed.

Amethyst/AppDelegate.swift
ianyh: I would treat it as an error if `focusedScreenManager` is nil. Maybe put a placeholder item in the submenu that just says it can't be found.

Amethyst/AppDelegate.swift
```diff
@@ -173,6 +184,66 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+
+            submenu.addItem(menuItem)
+        }
+
+        // If no layouts are enabled, show a disabled message
+        if enabledLayoutKeys.isEmpty {
+            let noLayoutsItem = NSMenuItem(title: "No layouts enabled", action: nil, keyEquivalent: "")
+            noLayoutsItem.isEnabled = false
+            submenu.addItem(noLayoutsItem)
+        }
```
ianyh: This is minor, but can you just do this check before the loop and then return if it's empty?

Amethyst/AppDelegate.swift
```diff
@@ -173,6 +184,66 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+        // Get current layout
+        let focusedScreenManager = windowManager?.focusedScreenManager()
+        var currentLayoutKey = focusedScreenManager?.currentLayout?.layoutKey
+
+        // If no focused screen manager, fallback to the first screen manager
+        if focusedScreenManager == nil, let firstScreenManager = windowManager?.screenManager(at: 0) {
+            currentLayoutKey = firstScreenManager.currentLayout?.layoutKey
+        }
+
+        // Filter to only enabled layouts and add menu items
```
ianyh: This logic isn't quite correct because it is technically possible for there to be multiple instances of the same layout in a space.

ianyh (changes requested):

Amethyst/AppDelegate.swift
rgeraskin: Fixed. Improved screen manager detection with a fallback chain:
1. Try `focusedScreenManager()` first
2. If nil (e.g., when clicking menu bar steals focus), detect the screen under the mouse cursor using `NSEvent.mouseLocation`
3. Fall back to `screenManager(at: 0)` as last resort
4. Show "Unable to determine current screen" placeholder only if all methods fail

This ensures the correct screen's layout is shown when clicking the menu bar on multi-monitor setups.

Amethyst/AppDelegate.swift
rgeraskin: Fixed. Changed the delegate from `layoutsMenuItem?.submenu` to `statusItemMenu`. The layouts menu is now populated in `menuWillOpen` when the main status item menu opens, making the list available before the user hovers over the Layouts submenu.

Amethyst/AppDelegate.swift
rgeraskin: Fixed. Moved the empty layouts check to the beginning of `populateLayoutsMenu()` with an early return.

Amethyst/AppDelegate.swift
rgeraskin: Fixed. Added `currentLayoutIndexValue` property to ScreenManager to expose the current layout index. The menu now compares by index instead of by layoutKey:

- Before: `currentLayout?.layoutKey == layoutKey` (would checkmark ALL instances of same layout type)
- After: `index == currentLayoutIndex` (only checkmarks the specific instance)

This correctly handles configurations like [Tall, Wide, Tall] where only the active instance shows a checkmark.

Amethyst/AppDelegate.swift
```diff
@@ -173,6 +177,83 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+            if let focused = windowManager?.focusedScreenManager() {
+                return focused
+            }
+            // Fallback to screen containing mouse cursor (useful when clicking menu bar)
+            let mouseLocation = NSEvent.mouseLocation
+            if let nsScreen = NSScreen.screens.first(where: { NSMouseInRect(mouseLocation, $0.frame, false) }) {
+                let amScreen = AMScreen(screen: nsScreen)
+                return windowManager?.screenManager(for: amScreen)
+            }
+            return windowManager?.screenManager(at: 0)
```
ianyh: I would just return `nil` here.

ianyh (changes requested):
My point about the layout instances is that we probably want this to be derived from the actual list of layouts in the screen manager.

Amethyst/AppDelegate.swift
rgeraskin: Removed the screenManager(at: 0) fallback. Now returns nil if both focusedScreenManager() and mouse cursor screen detection fail, which will show the error placeholder.

rgeraskin:
> My point about the layout instances is that we probably want this to be derived from the actual list of layouts in the screen manager.

Refactored to derive the menu from the screen manager's actual layouts instead of the global UserConfiguration.shared.layoutKeys():

- Added layoutsInfo property to ScreenManager that exposes the layouts array as [(key: String, name: String)]
- populateLayoutsMenu() now iterates over screenManager.layoutsInfo instead of global config
- This ensures indices match correctly since both menu items and currentLayoutIndex come from the same source

```
// Before: used global config (indices might not match screen manager)
let enabledLayoutKeys = UserConfiguration.shared.layoutKeys()
for (index, layoutKey) in enabledLayoutKeys.enumerated() { ... }
```

```
// After: derived from screen manager's actual layouts
let layouts = screenManager.layoutsInfo
for (index, layoutInfo) in layouts.enumerated() { ... }
```

Amethyst/Managers/ScreenManager.swift
```diff
@@ -56,6 +56,16 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
...
 
+    /// The index of the current layout in the layouts array
+    var currentLayoutIndexValue: Int {
+        return currentLayoutIndex
+    }
+
+    /// Returns layout info (key and name) for all layouts in this screen manager
+    var layoutsInfo: [(key: String, name: String)] {
+        return layouts.map { (key: $0.layoutKey, name: $0.layoutName) }
+    }
```
ianyh: * Could you change layoutsInfo to use a defined struct type instead of the named tuple?
* Could you include whether or not it is the selected layout in that struct instead of exposing the index separately above?

ianyh (changes requested):
One comment, but otherwise I think it looks good.

Amethyst/Managers/ScreenManager.swift
rgeraskin: done

ianyh (approved):

PR #1831, opened 2026-01-01
Fix issues with tab swapping and change distribution

ianyh:
- [x] Deliver change events to layouts on reflow to target correct layout
- [x] Observe window closes
- [x] Improve native tab handling
- [x] Make bsp state make sense

This fixes a variety of issues around state management. For one thing, window closes got lost at some point, and were not being delivered to stateful layouts (including custom ones). There was an additional issue around native tab swapping and what that actually means for layout state.

This closes many (but not all) issues with bsp.

Fixes #1614

PR #1839, opened 2026-02-15
Fix: Stabilize reflow during Space transitions and window tracking order

RyoZaizen:
## Summary
  1. In `4 Column Left`, window order kept changing over time even when no new windows were added.
  2. When switching Spaces via Dock to an app in a`Floating`-designated Space, an unintended rearrangement could run on arrival (windows shifting/reordering and overlapping).

  ## Problem Details
  - Reflow could run during Space transitions while the tracked Space state was stale.
  - Window tracking/swap/remove logic could become unstable due to equality checks based on window objects instead of stable IDs.
  - Potentially new windows could be processed again even if already tracked.

  ## Fixes
  - Run reflow only when `ScreenManager`'s tracked Space matches the actual current Space.
  - Skip frame assignment for windows that are inactive or floating.
  - Ignore newly discovered windows if they are already tracked.
  - Use window IDs for tracked/remove/swap comparisons to stabilize ordering.

 ### Follow-up
- [This guard](https://github.com/ianyh/Amethyst/pull/1839/changes/BASE..21db785d9f4d4427b8104b618799ede3078ce1a0#diff-e9290deac9df0562f5227e32a3d6e82b85b99238df9374798ad81cc6d976b3e4R597-R599) is a temporary mitigation and not the final fix.

### DIFF
--- a/Amethyst/Managers/ScreenManager.swift
+++ b/Amethyst/Managers/ScreenManager.swift
@@ -230,6 +230,12 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
             return
         }
 
+        // During rapid Space transitions, activation/focus notifications can arrive before
+        // this screen manager updates its tracked Space. Skip reflow if state is stale.
+        guard let currentSpace = CGSpacesInfo<Window>.currentSpaceForScreen(screen), currentSpace.id == space?.id else {
+            return
+        }
+
         guard let windows = delegate?.activeWindowSet(forScreenManager: self) else {
             return
         }
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -594,6 +594,10 @@ extension WindowManager: ApplicationObservationDelegate {
     }
 
     func application(_ application: AnyApplication<Application>, didFindPotentiallyNewWindow window: Window) {
+        guard !windows.isWindowTracked(window) else {
+            return
+        }
+
         swapInTab(window: window)
     }
 
--- a/Amethyst/Managers/Windows.swift
+++ b/Amethyst/Managers/Windows.swift
@@ -84,14 +84,14 @@ extension WindowManager {
         }
 
         func remove(window: Window) {
-            for (_, lastMainWindow) in lastMainWindows where lastMainWindow == window {
+            for (_, lastMainWindow) in lastMainWindows where lastMainWindow?.id() == window.id() {
                 if let currentFocusedSpace = CGSpacesInfo<Window>.currentFocusedSpace() {
                     let secondWindow = activeWindowOnCurrentScreen(atIndex: 1)
                     lastMainWindows[currentFocusedSpace.id] = secondWindow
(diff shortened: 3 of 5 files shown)


Amethyst/Managers/ScreenManager.swift
```diff
@@ -230,6 +230,12 @@ final class ScreenManager<Delegate: ScreenManagerDelegate>: NSObject, Codable {
             return
         }
 
+        // During rapid Space transitions, activation/focus notifications can arrive before
+        // this screen manager updates its tracked Space. Skip reflow if state is stale.
+        guard let currentSpace = CGSpacesInfo<Window>.currentSpaceForScreen(screen), currentSpace.id == space?.id else {
+            return
+        }
```
ianyh, line 237: Nice catch.

Amethyst/Managers/WindowManager.swift
```diff
@@ -594,6 +594,10 @@ extension WindowManager: ApplicationObservationDelegate {
     }
 
     func application(_ application: AnyApplication<Application>, didFindPotentiallyNewWindow window: Window) {
+        guard !windows.isWindowTracked(window) else {
+            return
+        }
```
ianyh, line 599: This isn't correct in all cases, which is something I'm working on in https://github.com/ianyh/Amethyst/pull/1839, but I might merge this for now because the tab bouncing is atrocious.

Amethyst/Managers/Windows.swift
```diff
@@ -84,14 +84,14 @@ extension WindowManager {
         }
 
         func remove(window: Window) {
-            for (_, lastMainWindow) in lastMainWindows where lastMainWindow == window {
+            for (_, lastMainWindow) in lastMainWindows where lastMainWindow?.id() == window.id() {
```
ianyh, line 87: I think I know what issue you were having here, which is also something I found in my testing. It seems like windows are sometimes in a different process that macOS is using for some window operations. That pid change would be the source of the equality not being correct. I'm hesitant to start doing it only by id, though, so could you try using this in pid instead? https://github.com/ianyh/Amethyst/pull/1831/changes#diff-7ab1a009ba3587b54018b4d381b528013286735661c149f18d4724ee5870abb7R251-R253
```swift
    func pid() -> pid_t {
        return forKey("AXParent" as CFString)?.processIdentifier() ?? processIdentifier()
    }
```

ianyh (changes requested):

Amethyst/Managers/WindowManager.swift
RyoZaizen: Thanks. I’ve added a follow-up note in the PR description.

Amethyst/Managers/Windows.swift
RyoZaizen: Thanks. done. [8e59296](https://github.com/ianyh/Amethyst/pull/1839/commits/8e59296e37bf7471fed3f326a22ff8e2190a470a)

ianyh (approved):

PR #1841, opened 2026-02-22
Amethyst 0.24.2

ianyh:


yangwao:
Does it fix some Tahoe 26.3 issues, like occasionally freezing window repositioning?

ianyh:
> Does it fix some Tahoe 26.3 issues, like occasionally freezing window repositioning?

I don't know that I have seen those issues.

Issue #1335, opened 2022-11-04
kitty does not tile


ChristianChiarulli:
**Describe the bug**
kitty does not respect tiling

**Applications:**
kitty Terminal

**To Reproduce**
Install kitty and try to tile

**Expected behavior**
Kitty should tile

**Versions:**
 - macOS: ventura 13.0
 - Amethyst: 0.16.0 (94)
 
**Debug Info**
```
Version: 0.16.0 (94)

OS version: Version 13.0 (Build 22A380)

Screens:
	(0.0, 0.0, 1470.0, 956.0) [(0.0, 0.0, 1470.0, 956.0)]

Configuration:
use-canary-build: 0
window-resize-step: 5
restore-layouts-on-launch: 1
ignore-menu-bar: 1
debug-layout-info: 0
enables-layout-hud: 0
screen-padding-left: 0
follow-space-thrown-windows: 1
focus-follows-mouse: 0
float-small-windows: 0
floating: (
        {
        id = "org.blenderfoundation.blender";
        "window-titles" =         (
        );
    },
        {
        id = "com.hnc.Discord";
        "window-titles" =         (
        );
    },
        {
        id = "org.gimp.gimp-2.10:";
        "window-titles" =         (
        );
    },
        {
        id = "im.riot.app";
        "window-titles" =         (
        );
    },
        {
        id = "com.apple.systempreferences";
        "window-titles" =         (
        );
    },
        {
        id = "com.valvesoftware.steam";
        "window-titles" =         (
        );
    }
)
enables-layout-hud-on-space-change: 0
new-windows-to-main: 1
window-minimum-height: 0
floating-is-blacklist: 1
screen-padding-right: 0
smart-window-margins: 1
mod1: (
    option,
    shift
)
screen-padding-bottom: 0
screen-padding-top: 0
window-margin-size: 10
mod2: (
    option,
    shift,
    control
)
layouts: (
    tall,
    fullscreen
)
mouse-follows-focus: 0
window-margins: 1
window-minimum-width: 0
```

**Additional context**
Was working fine before ventura update

paulmiddelkoop:
I have the same problem. For me, most of the times the first Kitty window is tiled, but the second one seems to float.

jammus:
I have the same problem but am still on Monterey

jammus:
Trying to debug this further...

* I'm using Kitty installed via Nix (https://github.com/jammus/dotfiles)
* I'm generally launching kitty using skhd
[...]

PR #1870, opened 2026-07-04
Fix tiling after reopening apps that keep windows alive

mmv08:
## Summary

Fixes a window-tracking issue where apps such as Telegram, Slack, Claude, and potentially Kitty can stop tiling after a window is closed or hidden, then later shown again.

Some apps keep the same underlying window alive instead of destroying it. When the window is shown again, Amethyst can receive a new accessibility element for the same `(pid, CGWindowID)` while still tracking the old element. The old element still matches by identity, but reads such as frame, title, and screen can fail. As a result, the live element is rejected as already tracked, and the window is not laid out.

This should theoretically address [ianyh/Amethyst#1335](https://github.com/ianyh/Amethyst/issues/1335). That thread includes a Telegram report with the same visible behaviour, and the kitty workaround of setting `macos_quit_when_last_window_closed yes` points at the same class of lifecycle issue: forcing the app to quit avoids reusing a hidden/retained window after the last window is closed. I have not reproduced the kitty case locally, so I am not treating that issue as definitively fixed.

## Changes

- Refresh the stored tracked window when a live accessibility element arrives for the same window identity.
- Preserve window order and floating state while replacing the stale element.
- Move the window-destroyed observer from the old element to the new element.
- Regenerate the active window ID cache and reflow the affected screen after a refresh.
- Fix `Windows.replace(window:withWindow:)` so it replaces the first argument with the second, matching the method name and tab detection call sites.
- Log window tracking failures instead of silently swallowing failed subscriptions.

## Disclosure

This bug had been annoying me for a while, and I used it as a chance to try Claude Fable 5 while debugging and preparing the fix. I am not an expert in macOS accessibility or window-server internals, but I walked through the behaviour, the fix makes sense to me, and I tested it locally.

## Testing

- Ran the full Amethyst test suite.
- Manually verified with Telegram that reopening the window returns it to the layout.
- Verified that `mod+z` can recover the window after reopening.

## Known limitation

Closing one of these hidden windows still may not immediately reflow the remaining windows, because macOS does not appear to send an accessibility notification for that hide transition. This PR focuses on adopting the live window when it reappears.

### DIFF
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -450,6 +450,7 @@ extension WindowManager {
 
         guard !windows.isWindowTracked(window) else {
             log.debug("Window was already tracked: \(window)")
+            refreshTrackedWindow(window, application: application)
             return
         }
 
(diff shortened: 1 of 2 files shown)


mmv08:
video of the app with and without the fix

https://github.com/user-attachments/assets/f7394d0c-3633-4500-90da-76324b9d1743

mmv08:
Follow-up: one of my machines never had this bug, and I couldn't figure out why. Turns out the bug only happens when macOS doesn't deliver `kAXUIElementDestroyedNotification` on window hide. When the notification arrives, Amethyst untracks the window the moment it hides and adopts the revived one as new, so nothing breaks.

On my affected machine, the culprit was Grammarly Desktop: while it's running, macOS stops delivering that notification at all. Quit it and even prod Amethyst without this fix works fine. Other AX apps are known to cause the same thing (koekeishiya/yabai#2431: Contexts, Phoenix, Amazon Q).

So Grammarly was the trigger, but with this fix, Amethyst doesn't depend on that notification at all, so it's immune either way.

Issue #1398, opened 2023-02-02
Some Applications Don't Tile Randomly


landoncrabtree:
**Describe the bug**
There seems to be an issue with some applications getting tiled properly. Sometimes they will tile properly, and other times they just launch like a regular application, ignoring the tiling layout. I believe it seems to be an issue with identifying if the window is floating or not. 

**Applications:**
Amethyst, Terminal

**To Reproduce**
Tile applications (ie: Firefox)
Open Terminal - Sometimes it will tile, othertimes it will not. Seems to be random. 

**Expected behavior**
Application should tile everytime. 

**Screenshots**
https://gyazo.com/fdc18b1fa7cfac49efa3e7a7a3fb29b5

**Versions:**
 - macOS:  13.1 (22C65) 
 - Amethyst: 0.17.0 (97)
 
**Debug Info**
```
$ /Applications/Amethyst.app/Contents/MacOS/Amethyst --debug-info [--include-apps]
```
Note: `--include-apps` will list your manageable applications, but is optional if you don't want to list that.
```
Version: 0.17.0 (97)

OS version: Version 13.1 (Build 22C65)

Screens:
	(0.0, 0.0, 4096.0, 1152.0) [(0.0, 0.0, 4096.0, 1152.0)]

Configuration:
screen-padding-bottom: 0
floating-is-blacklist: 1
mouse-resizes-windows: 1
focus-follows-mouse: 0
use-canary-build: 0
new-windows-to-main: 0
screen-padding-right: 0
mod2: (
    option,
    shift,
    control
)
ignore-menu-bar: 0
debug-layout-info: 0
follow-space-thrown-windows: 1
screen-padding-top: 0
floating: (
)
window-max-count: 0
window-margin-size: 0
screen-padding-left: 0
window-resize-step: 5
float-small-windows: 1
mouse-follows-focus: 0
enables-layout-hud: 1
layouts: (
    tall,
    wide,
    fullscreen,
    column,
    "3column-right",
    "middle-wide"
)
enables-layout-hud-on-space-change: 1
window-minimum-height: 0
window-margins: 0
window-minimum-width: 0
mouse-swaps-windows: 0
restore-layouts-on-launch: 1
mod1: (
    option,
    shift
)
```

**Additional context**

jaming715:
Same issue here

TomEstelrich:
Same happens to me randomly

gilitos92:
Same issue here

[...]

PR #1871, opened 2026-07-08
Scan for untracked windows after tab close

cblecker:
- When closing the active tab in apps with native macOS tabs (e.g. Ghostty), the newly revealed tab may never have been tracked, and the app may not fire a notification Amethyst observes for it. Amethyst loses track of the remaining window, leaving a hole in the layout.
- After removing a destroyed window, scan the same application for windows that are now on screen and untracked, and add them.
- This is a targeted mini-reevaluate scoped to just the affected application.

How the scan is kept narrow:
- Element destruction gets its own `didDestroyWindow` delegate callback, so minimizing a window (which also routes through `didRemoveWindow`) does not trigger the scan or re-track the minimized window.
- The application's window cache is dropped before scanning, matching the existing scans on unhide and space change.
- Candidates are filtered to windows other than the destroyed one that are active and on screen (`isWindowActive`), then to untracked ones, before `add(window:)` runs. That costs a few cheap attribute reads per window in the app; the full add path only runs for on-screen windows that aren't tracked, which in the common case is just the surviving tab.
- Any pending tab-detection state for the survivor is cleared before adding it, so a late focus event cannot run tab detection against it and swap it for a stale window.

Related issues: #1335 (kitty doesn't tile — same class of native-tab lifecycle issue), #1398 (apps randomly don't tile)

### DIFF
--- a/Amethyst/Managers/WindowManager.swift
+++ b/Amethyst/Managers/WindowManager.swift
@@ -773,6 +773,25 @@ extension WindowManager: ApplicationObservationDelegate {
         remove(window: window)
     }
 
+    func application(_ application: AnyApplication<Application>, didDestroyWindow window: Window) {
+        remove(window: window)
+
+        // In apps with native macOS tabs, each tab is its own window and only the visible
+        // one is tracked. Closing the active tab destroys the tracked window and reveals a
+        // sibling that may never have been tracked, and the app may fire no notification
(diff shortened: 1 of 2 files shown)


cblecker:
Hi @ianyh ! Sorry for the direct ping -- just want to understand if there is anything else I should do to nudge this one forward, or if it's just pending on you having time to take a look. Thanks for this great application (I've used it daily for years!)

